PR #74, opened 2018-09-14
Command, Command Handling, Aggregate Roots and Values/Concepts documentation

TomasEkeli:


Documentation/articles/command/command_handler.md
```diff
@@ -0,0 +1,44 @@
...
+## Background
+
+A *Command Handler* handles commands.  The command handler is responsible for executing the appropriate methods on Aggregate Roots and Domain Services to fulfill the intent of the *Command*.
+
+## Requirements
+
+A *Command Handler* must implement the *IHandleCommands* interface.  This is a simple marker interface with no methods to implement.
+
+To handle a *Command*, you create a method called *Handle* with a single parameter that is the *Command* you wish to handle.  A *Command Handler* does not return anything.  It is not permitted to return a value from a *Command*.  When a *Command* is executed, you will receive a *CommandResult*
+which will indicate whether the *Command* succeeded and, in the case of failure, why the *Command* failed.  No artifacts that were created or updated as a result of the *Command* will be present on the *CommandResult*.
```
smithmx: We shouldn't use the term "artifact" here, we are using that for another purpose now and it could be confusing.

Documentation/articles/command/command_handler.md
```diff
@@ -0,0 +1,44 @@
...
+
+```csharp
+public void Handle(AddRecommendationToCart cmd)
+```
+
+There **MUST** be a single handler for a Command.  Dolittle will throw an exception if there is no handler for a command or if there is more than one handler for a command.  There can only be a single handler for a command as there must be a clear and unambiguous indication of the result of the execution of the command.  Multiple command handlers for a single command would require co-ordination of multiple results with associated rollback of successful handlers after unsuccessful ones.  Where you wish multiple things to happen on receipt of a Command, you must implement and handle this yourself in your *Command Handler*.
+
```
smithmx: We'll probably want to re-inforce this by referring to the (unwritten) Event Store documentation that says that a commit is against a single event source.  Indeed, the whole thing is about a transaction.  A command is a transaction.  Therefore the command handler must be a transaction.  Since a transaction can only be a single commit and the Event Store only supports a commit against a single event source, then a command handler can only get a single event source.

This is more just a note to myself that we need to flesh this out a bit later! :-)

Documentation/articles/command/command_handler.md
```diff
@@ -0,0 +1,44 @@
...
+The sole responsibility of the *Command Handler* is to execute the *Command*.
+
+A *Command Handler* should not implement any validation, security checks, or such like.  These should be handled earlier in the command pipeline by validators.  When we invoke a *Command Handler*, it is with the expectation that the action will succeed.  We should have validated the command,
+checked business rule conditions, checked the permissions and therefore the command should succeed.  It is for this reason that within a *Command Handler*, the only way to indicate that the *Command* did not succeed is via an Exception.  This is because, due to our previous checks,
+it is truly an exceptional condition for the *Command* to fail.
+
+## Design
+
+It is recommended that a *Command Handler* should invoke a single method on an *Aggregate Root*.  This is an indication that you have correctly modelled your *Aggregate Root*, *Command* and business transaction.  It is sometimes useful to think of a *Command* as a serialized call on a method on an *Aggregate Root*. If you are calling multiple *Aggregate Roots* you will need to handle one of them failing and rolling back the transaction.
```
smithmx: Refer to the comment above.  This will actually not work going forward.  Bifrost supported tracking multiple aggregates but Dolittle won't... because of the Event Store / Commit metaphor.

Documentation/articles/command/command_handler.md
```diff
@@ -0,0 +1,44 @@
...
+
+{{% notice tip %}}
+Calling multiple *Aggregate Roots* from the same command handler for the same commands indicates that you may need to create a new *Aggregate Root* that captures this business transaction as one thing. It can be fruitful to consider modelling the verb, not the noun (i.e. the adding of a product
+to the cart, not the cart with an add -method).
```
smithmx: I really need to learn to read ahead before commenting... :-)

Documentation/articles/command/command_handler.md
```diff
@@ -0,0 +1,44 @@
...
+{{% /notice %}}
+
+As nothing can be returned from a *Command Handler*, we discourage reliance on database-generated identities or keys. You will not be able to receive the Id generated by the database. It is recommended to use non-meaningful keys such as Guids on a *Command* which can be provided by the invoker of the *Command* and *Command Handler*. Where a database generated key is required, it is recommended to include another candidate key that can be provided by the caller.
```
smithmx: Believe it or not but we've encountered opposition to this on the basis of the cost of having a guid per row...

Documentation/articles/command/introduction.md
```diff
@@ -0,0 +1,44 @@
...
+
+A command **MUST** include all necessary information to perform the action.  These **SHOULD** be in the form of parameters on the command object. You **MAY** include optional parameters, though it is **recommended** that you create multiple commands that represent the different states associated with the optional parameters.  It is **RECOMMENDED** that you use [Concepts](../concepts_and_value_objects.md) and [Value Objects](../concepts_and_value_objects.md) on your commands rather than primitives. This gives a more expressive command and aids in validation.
+
+{{% notice tip %}}
+> For JavaScript, proxy representations can be used for the commands, read more about the mechanism [here](../../Frontend/JavaScript/proxy_generation.md).
+{{% /notice %}}
+
+## Relation to Events
+
+An important point about commands is that while it is an imperative, it can fail. Once the command has clears validation it will cause actions to happen in the system that lead to events. These capture the changes in the system and cannot fail.
```
smithmx: "has cleared" or "clears"

Documentation/articles/domain_driven_design/aggregate_root.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Background
+
+An *Aggregate* is an object or collection of objects (Entities and Value Objects) that must be treated as a single unit for the
+purposes of a business transaction. The *Aggregate* will maintain all invariants for all objects that comprise the *Aggregate*,
+enforcing all business rules internal to the *Aggregate*.
+
+# Aggregate Root 
+
+All *Aggregates* have a single point of entry that is know as the root.  The *Aggregate Root* is the interface to the external world.
```
smithmx: These have to be Guids, we don't support anything else, we should be upfront about that here.

Documentation/articles/domain_driven_design/aggregate_root.md
```diff
@@ -0,0 +1,64 @@
...
+All interaction with an *Aggregate* is via the *Aggregate Root*.  As such, an *Aggregate Root* **MUST** have a globally unique identifier
```
smithmx: Hey, just put "(GUID)" in brackets after globally unique identifier and that takes care of the comment above! :-)

Documentation/articles/domain_driven_design/aggregate_root.md
```diff
@@ -0,0 +1,64 @@
...
+
+The *Aggregate* maintains only its internal invariants. It is not responsible for maintaining invariants between *Aggregates*.
+
+For example, it is not the responsiblity of the individual *Aggregate* to ensure that a *Username* is unique.  It is the responsibility of the **System as a whole** to maintain these system invariants.  The exact mechanism it uses will depend on the invariant and the consequences of breaking them.
+
+{{% notice tip %}}  
+> What you must learn is that these rules are no different that the rules of a computer system. Some of them can be bent. Others can be broken.
+> <cite>Morpheus, The Matrix</cite>
+
+All rules are not created equal.  Business Rules are essentially business problems, not primarily technical problems.  It may be simpler and cheaper
```
smithmx: Udi Dahan has a great example of this.  He's talking about an ATM machine and how it is not absoutely 100% transactionally consistent at all times on the account balance.  So in certain circumstances it's actually possible to withdraw more than your balance.  As devs we look at that as a problem to be solved and can dream up ridiculously complex and expensive solutions, while the business see it as an opportunity... unauthorized overdrafts are expensive and a good source of profit for next to nothing for the bank!

Documentation/articles/domain_driven_design/aggregate_root.md
```diff
@@ -0,0 +1,64 @@
...
+
+## State
+
+In Dolittle, an *Aggregate* is a purely *WRITE* concern. The aggregate plays no role in reading or presentation. As such, an *Aggregate* **MUST NOT** expose any public state, via variables or properties. Neither should the *Aggregate* return any of its internal entities or values objects, even in a transient state. The *Aggregate* is only required to keep such internal state that it requires to maintain its invariants.  Since no public state is exposed by the *Aggregate* it does not need to keep in state values that are not part of its internal structure. 
+
+This is the natural conclusion of all actions on the *Aggregate* springing from the *Aggregate Root*. If the root were to allow access to its internal entities or value-objects they could be stored past the lifetime of the *Aggregate Root* or even changed without running through the gatekeeper that is the root.
+
+## Identity
+
+All *AggregateRoot* classes must have a *Guid* key. It is permissible to include other candidate keys, though these should have a map to the Guid.
```
smithmx: Forget my other two comments on this issue! :-)

Documentation/articles/domain_driven_design/concepts_and_value_objects.md
```diff
@@ -0,0 +1,114 @@
...
+While ultimately all concepts must be expressed in the primitive types available in the language writing
+the domain-code like this leads to a less expressive, less correct domain model. For example, a telephone
+number might be represented as a *string* (or even an integer).  However, there is logic associated with
+a telephone number and aspects of it that are not necessarily captured by a string.  We may be interested
+in area codes, international dialling codes, extensions, country specific formatting rules, and so on.
+Replacing a primitive with a type allows us to capture that while not all strings are telephone numbers, 
+all telephone numbers can be expressed as strings.
+
+{{% notice warning %}}
+Never EVER represent things that look like "numbers" by virtue of being composed solely of digits as integers.
```
smithmx: ** MUST NOT ** ?

Documentation/articles/domain_driven_design/concepts_and_value_objects.md
```diff
@@ -0,0 +1,114 @@
...
+{{% /notice %}} 
+
+# Value Objects in Dolittle
+
+Value object **MUST** be treated as immutable after creation and initial population.  Dolittle does not
```
smithmx: We might actually be in a position to enforce this now.  But it's correct as it currently stands...

Documentation/articles/domain_driven_design/concepts_and_value_objects.md
```diff
@@ -0,0 +1,114 @@
...
+    public static implicit operator TelephoneNumber(string value)
+    {
+        return new TelephoneNumber { Value = value };
+    }
+}
+```
+
+While the underlying ConceptAs<T> class can handle the implicit conversion to a string, it is not possible
+to implement the conversion to the specific type (i.e. TelephoneNumber) in a generic manner in the base class.
+
```
smithmx: Concept<T> now supports IComparable<Concept<T>> and IComparable and has also overloaded the operators < > <= >=.  This makes it much easier to replace the primitive with the concept.

Documentation/articles/domain_driven_design/concepts_and_value_objects.md
```diff
@@ -0,0 +1,114 @@
...
+It is *RECOMMENDED* and *ENCOURAGED* to use *ConceptAs\<T>* as properties within *Value\<T>*.  It is 
+*DISCOURAGED* but *NOT PROHIBITED* from using *Value\<T>* with just a single property.
+
+{{% notice tip %}}  
+Value objects within Dolittle consider the *System Type* to be integral to their identity.  While it is 
+permissible to inherit from a Value\<T> or ConceptAs\<T> that you have created, a type of the base class and 
+a type of subclass **WILL NOT** be considered equal, even if they have all properties identical.  A value
+object can only be equal to itself or another object of the same type.
+{{% /notice %}}  
+
```
smithmx: There's no IComparable<> support for Value<>.  We have no way of knowing that so the user has to do it themselves.

Documentation/articles/command/command_handler.md
TomasEkeli: amazing

PR #76, opened 2018-09-16
Documentation: Querying for Data, Queries & Read Models

pavsaund:
Opening a PR for documentation for the read-side of things.

Querying for Data - article explaining the how behind querying for data
Queries - Explanation of Queries in Dolittle 
Read Models - Explanation of ReadModel in Dolittle

This is still a work in progress and not ready to be merged, but I'd like some feedback on the article.

### DIFF
--- a/Documentation/Articles/querying_for_data.md
+++ b/Documentation/Articles/querying_for_data.md
@@ -14,8 +14,8 @@ The query or read side of the CQRS pattern is all about exposing the current sta
 {{% notice note %}}
 Dolittle exposes read data through two different mechanisms:
  
- - [Read Models](https://github.com/dolittle/Runtime/tree/master/Documentation/read/read_models.md)
- - [Queries](https://github.com/dolittle/Runtime/tree/master/Documentation/read/queries.md)
+ - [Read Models](https://github.com/dolittle/Runtime/tree/master/Documentation/read/read_model.md)
+ - [Queries](https://github.com/dolittle/Runtime/tree/master/Documentation/read/query.md)
 {{% /notice %}}
 
 ### Updating Read Data
--- a/Documentation/Articles/querying_for_data.md
+++ b/Documentation/Articles/querying_for_data.md
@@ -1,0 +1,89 @@
+---
+title: Querying for Data
+description: About Reading data
+keywords: read model, query, ddd, read, cqrs,  
+author: pavsaund
+---
+
+
+# Querying for Data
+
+## Introduction
+The query or read side of the CQRS pattern is all about exposing the current state of a system in a format that is ready to be consumed by interaction layers (ie: web-based user interface). The state should be exposed through models that are optimized for querying, and is usually updated by changes made to a system through events.
+
+{{% notice note %}}
+Dolittle exposes read data through two different mechanisms:
+ 
+ - [Read Models](https://github.com/dolittle/Runtime/tree/master/Documentation/read/read_models.md)
+ - [Queries](https://github.com/dolittle/Runtime/tree/master/Documentation/read/queries.md)
+{{% /notice %}}
+
+### Updating Read Data
+[Event Processors](LINK TO EVENT PROCESSORS) are responsible for consuming events in the system and updating read-models.
+
+### Storing Data 
+#### Optimize for querying
+Unlike traditional relational data models that are spread across tables, read data should be optimized as much as possible to the clients consuming that data. This means creating models that can be consumed directly without additional queries. One of the of the benefits is that reading data becomes lightweight, fast and allows for good user experiences.
+
+Duplicating data in 
+
+{{% notice tip %}}
+Think of data on the query side as your applications cache. It get's updated whenever there's a change to the system, and is fast by default.
+{{% /notice %}}
+
+##### Example of optimizing for querying:
+In an e-commerce solution, you get a requirement to show a small-cart summary with Number of Items and total price. This is a challenge since it will get the current ShoppingCartModel each time the user hits any page of the site, and it adds a visible delay to the user.
+
+A possible workaround could be to introduce a second cart model that only cared about the small-cart information and had pre-calculated toals.
+
+Full ShoppingCartModel:
+```
+ShoppingCartModel
+    |-CartId
+    |-CartItems[]
+        |-ProductName
+        |-Price
+        |-TaxRate
+        |-Quantity
+        |-Dimensions
+        |-...
+    |-AdditionalServices[]
+        |-ServiceType
+        |-Price
+        |-Quantity
+        |-...
+    |-GiftCards[]
+        |-...
+    |-VoucherCodes[]
+    |-Totals
+        |-VatRate
+        |-Net
+        |-Gross
+```
+
+Focused SmallCartModel
+```
+SmallCartModel
+    |-CartId
+    |-ItemsCount
+    |-Totals
+        |-VatRate
+        |-Net
+        |-Gross
+```
+
+Now with the new model, queries can be made for the SmallCartModel on every page load, instead of the larger ShoppingCartModel.
+
+
+
+#### Adapt storage
+Another benefit is that read data can be persisted in the most suited format and not limited to any other storage mechanisms. Examples could be: Document Databases, JSON-files, XML-files, Relational DB, In-memory, Excel / PDF files etc. 
+
+{{% notice tip %}}
+Use the storage mechanism that most suits your use cases for reading data like generating a json file persisted to a CDN ready to be consumed by a web-frontend.
+{{% /notice %}}
+
+
+## Eventual consistency / Asynchronous
+
+
--- a/Documentation/read/query.md
+++ b/Documentation/read/query.md
@@ -9,10 +9,14 @@ author: einari, smithmx
 Looking up collections of data to be presented to users in a view is a common scenario in applications. Dolittle formalizes this through `Query`-types. These types become the contract for your query capabilities. Queries can be optimized for the datasource as you see fit for the feature they're being used in.
 
 ## QueryFor
-To implement a query, you need to implement the interface `IQueryFor<>` with the generic type argument of a [read model](read_model.md). By convention the query expects a public property with a getter named `Query` with a return type of `IQueryable<>`., which can be expanded based on your specific needs. 
-returns is recognized by Bifrost and Bifrost will look for a [query provider](query_providers.md)
+To implement a query, you basically need to implement the interface `IQueryFor<>`.
+The generic type argument should point to a [read model](read_model.md).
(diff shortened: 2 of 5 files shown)


Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Querying for Data
+description: About Reading data
+keywords: read model, query, ddd, read, cqrs,  
+author: pavsaund
+---
+
+
+# Querying for Data
```
smithmx: I'd prefer something like "Querying your System" rather than data.

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Introduction
+The query or read side of the CQRS pattern is all about exposing the current state of a system in a format that is ready to be consumed by interaction layers (ie: web-based user interface). The state should be exposed through models that are optimized for querying, and is usually updated by changes made to a system through events.
```
smithmx: Maybe something to emphasise that the read / query side can be whatever you want it to be: you can and will use a number of different models / technologies.

Off the top of my head:

The query or read side of the CQRS pattern is all about exposing the current state of your system in a format that is optimized for particular requirements.  Since there are competing and conflicting requirements, no single data storage technology or data format will be optimal for all scenarios.  It is therefore to be expected that in a sufficiently complex system there will be polyglot persistence and querying.  A typical example, where the main requirement is to display information to an end user, might involve persistence to a document database or relational database.  Alternatively, querying might involve populating a search index or nodes in a graph database. 

Since state is updated through event subscriptions, we can store the state in multiple forms and for multiple purposes, with a query optimized for a specific purpose.

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+
+{{% notice note %}}
+Dolittle exposes read data through two different mechanisms:
```
smithmx: To help with common scenarios such as presentation of data to end users, decision supports, etc., Dolittle provides two helpers for retreiving data:

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+ 
+ - [Read Models](https://github.com/dolittle/Runtime/tree/master/Documentation/read/read_model.md)
+ - [Queries](https://github.com/dolittle/Runtime/tree/master/Documentation/read/query.md)
+{{% /notice %}}
+
+### Updating Read Data
+[Event Processors](LINK TO EVENT PROCESSORS) are responsible for consuming events in the system and updating read-models.
```
smithmx: I would say "subscribing to events" and updating read-models when state changes.

Documentation/read/query.md
```diff
@@ -0,0 +1,175 @@
...
+    {
+        // Require that hiring date is set after a certain date
+        ForArgument(e => e.HiredAfter).HasToBeGreaterThan(new DateTime(1985,1,1));
+    }
+}
+```
+
+## Paging
+
+Paging is not something you need to think about. This is what the [query provider](query_provider.md)
```
smithmx: The QueryProvider model for IQueryable only works with paging if there is a Linq provider that supports deferred execution and no-one in the chain has called ToList or ToArray.

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Storing Data 
+#### Optimize for querying
+Unlike traditional relational data models that are spread across tables, read data should be optimized as much as possible to the clients consuming that data. This means creating models that can be consumed directly without additional queries. One of the of the benefits is that reading data becomes lightweight, fast and allows for good user experiences.
+
+Duplicating data in
```
TomasEkeli: This just peters out - is something missing?

Something like "Duplicating data in several places to optimise for reads is considered a good thing when using Dolittle. It is the responsibility of the Command-side to keep these optimised structures up-to-date and correct. The Read-side merely gives the data it has access to with minimal logic and fuss. This might seem foreign and weird, but it simplifies both the read and write sides a lot, and lets systems be responsive and performant on the read-side independently of the write-side."

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+#### Optimize for querying
+Unlike traditional relational data models that are spread across tables, read data should be optimized as much as possible to the clients consuming that data. This means creating models that can be consumed directly without additional queries. One of the of the benefits is that reading data becomes lightweight, fast and allows for good user experiences.
+
+Duplicating data in 
+
+{{% notice tip %}}
+Think of data on the query side as your applications cache. It get's updated whenever there's a change to the system, and is fast by default.
+{{% /notice %}}
+
+##### Example of optimizing for querying:
```
TomasEkeli: Could change to something like "Write specific read-models for specific needs"

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+In an e-commerce solution, you get a requirement to show a small-cart summary with Number of Items and total price. This is a challenge since it will get the current ShoppingCartModel each time the user hits any page of the site, and it adds a visible delay to the user.
```
TomasEkeli: I'd emphasise having a fit-for-purpose model over the full ShoppingCartModel being slow (as it shouldn't be when it's pre-calculated, excepting bandwidth considerations). The point is to have a model that slots nicely into the need of the view.

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+        |-Net
+        |-Gross
+```
+
+Now with the new model, queries can be made for the SmallCartModel on every page load, instead of the larger ShoppingCartModel.
+
+
+
+#### Adapt storage
+Another benefit is that read data can be persisted in the most suited format and not limited to any other storage mechanisms. Examples could be: Document Databases, JSON-files, XML-files, Relational DB, In-memory, Excel / PDF files etc.
```
TomasEkeli: or even fully rendered html documents (perhaps with JSON models inlined when interaction is needed) - pre-computed and available just in case the page is ever needed.

Documentation/Articles/querying_for_data.md
```diff
@@ -0,0 +1,89 @@
...
+#### Adapt storage
+Another benefit is that read data can be persisted in the most suited format and not limited to any other storage mechanisms. Examples could be: Document Databases, JSON-files, XML-files, Relational DB, In-memory, Excel / PDF files etc. 
+
+{{% notice tip %}}
+Use the storage mechanism that most suits your use cases for reading data like generating a json file persisted to a CDN ready to be consumed by a web-frontend.
+{{% /notice %}}
+
+
+## Eventual consistency / Asynchronous
+
```
TomasEkeli: this needs a bit more ;-)

Documentation/read/query.md
```diff
@@ -0,0 +1,175 @@
...
+---
+
+# Query
+Looking up collections of data to be presented to users in a view is a common scenario in applications. Dolittle formalizes this through `Query`-types. These types become the contract for your query capabilities. Queries can be optimized for the datasource as you see fit for the feature they're being used in.
+
+## QueryFor
+To implement a query, you basically need to implement the interface `IQueryFor<>`.
+The generic type argument should point to a [read model](read_model.md).
+This is a marker interface and does not require you implement anything specific.
+Instead it is relying on a convention; its looking for a property called `Query`.
```
TomasEkeli: The property must be public, and the return-type must be (at least in the version I'm using) IQueryable<T> where T : IReadModel

Documentation/read/query.md
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+```
+
+## Paging
+
+Paging is not something you need to think about. This is what the [query provider](query_provider.md)
+deals with. Paging is a frontend concern and should not something you need to think about.
+In fact, paging is ideally a view concern, something decided as part of the interaction design.
+The client representations of queries has functionality related to paging.
```
TomasEkeli: I disagree. Paging cannot really be solved in a general way, and should be implemented as proper arguments to the Query. It places large restrictions on the interaction to the underlying data-storage to support paging in general, and having paging-like arguments that make sense in the language of the domain is better anyway (i.e. on other things than page-size and page-number, like from/to dates or highest-priority-issues). To really have paging based on "pages" you'd need to "lock-in" the read-side when you send the first query to avoid repeats across pages or missing read-models because they fell between pages.

Documentation/read/query.md
```diff
@@ -0,0 +1,175 @@
...
+    public IEnumerable<IReadModel> Filter(IEnumerable<IReadModel> readModels)
+    {
+        var result = new List<IReadModel>();
+        foreach( var readModel in readModels )
+        {
+            if( readModel is Employee && ((Employee)readModel).HiredAfter > new DateTime(1985,1,1) )
+                result.Add(readModel);
+        }
+        return result;
+    }
```
TomasEkeli: `return readModels
    .Where(rm => rm is Employee)
    .Where<Employee>(rm => rm.HiredAfter >= new DateTime(...));`

Documentation/read/query.md
```diff
@@ -0,0 +1,175 @@
...
+## QueryResult
+
+At the top of the pipeline sits the `QueryCoordinator`, this will execute the query and return a
+`QueryResult`object.
+
+| Property         | Type                    | Description                                         |
+| ---------------- | ----------------------- | --------------------------------------------------- |
+| Items            | `IEnumerable`           | Contains the items                                  |
+| TotalItems       | `Integer`               | Total number of items returned                      |
+| QueryName        | `Boolean`               | Name of the query                                   |
```
TomasEkeli: not a boolean

TomasEkeli: string, not boolean

Documentation/Articles/querying_for_data.md
pavsaund: Resolving as I go here - not sure if this is the way to do things.

Documentation/Articles/querying_for_data.md
pavsaund: The intention here is to link a reader to the "previous" step of actually generating the data from events. I can remove it to keep the document more focused

Documentation/Articles/querying_for_data.md
pavsaund: oh, wow - I must have dozed off while writing this.

Documentation/Articles/querying_for_data.md
pavsaund: Leaving this part open - mostly because I'm not sure what to write here.

PR #91, opened 2018-10-10
Resource system

woksin:


einari:
Let's hold off on the [Fundamentals PR](https://github.com/dolittle/DotNET.Fundamentals/pull/98) to go through with all specs in place being all green and sweet.

PR #100, opened 2018-10-30
Ignore registering event processors  for system and unknown tenants

woksin:


woksin:
Not creating eventStore_for_unknown/system anymore 

screenshot 2018-10-30 at 08 54 58

Source/Events/Processing/BootProcedure.cs
```diff
@@ -85,8 +86,11 @@ void ProcessInParallel()
                 {
                     Parallel.ForEach(_tenants.All, (t) =>
```
einari: I don't think this code should know about these. The _tenants.All shouldn't return System and Unknown

woksin: Makes sense

PR #101, opened 2018-11-02
Document required convention in events

TomasEkeli:
This tripped me up, so we should at least document it.


woksin:
I think that we can ignore travis build

PR #115, opened 2018-11-13
Event horizon should work now + moved Application configuration stuff to Runtime

woksin:


Source/Events/Processing/ScopedEventProcessor.cs
```diff
@@ -71,16 +71,16 @@ public virtual void CatchUp()
         {
             using(var repository = _getOffsetRepository())
             {
-                LastVersionProcessed = _getOffsetRepository().Get(ProcessorId);
+                LastVersionProcessed = repository.Get(ProcessorId);
             }
-
-            SingleEventTypeEventStream eventStream = _getUnprocessedEventsFetcher().GetUnprocessedEvents(Key.Event.Id,LastVersionProcessed) ?? new SingleEventTypeEventStream(null);
+            var unprocessedEventsFetcher = _getUnprocessedEventsFetcher();
```
smithmx: Perhaps IFetchUprocessedEvents should be IDisposable.  It's not really needed as the **implementation** uses the EventStore internally with a factory func and a using statement but it would read more clearly and cause less confusion it if followed the same pattern as the repository directly above it.

This isn't directly related to this commit of course...

PR #135, opened 2019-01-07
Making it possible to disable gRPC server

einari:


einari:
No. It has a default value of enabled - which is correct.

einari:
@smithmx This pull request now became two things; they both have to go in. But I didn't have this particular change in a branch and therefore it got latched on to this. 

The additional things in addition to the configurable /enabling-disabling of gRPC server is that we are now making our Query pipeline Task<> based. It supports synchronous and asynchronous queries in the implementation, so it won't break any existing queries implemented. This is a pre-requisite for getting things working in WebAssembly with MiniMongo, which is Promise continuation based. Been desperately trying to come up with completely synchronized versions - but always ends up halting and locking the thread. 

This affects primarily our interaction end points, which is WebAssembly - which is new, and our AspNetCore project. It does not affect any client code calling into it. So we have to look at how we version this particular change before we publish.

PR #259, opened 2019-06-21
Plugins

woksin:
Adds the runtime tooling plugin and boilerplates

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1115616103797671/1128103608714310)


Boilerplates/Application/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
einari: If I'm reading this right, every Application created will have this license - we can't enforce that. So we should therefor delete this file

woksin: Nope, only the stuff in the Content folder is created by the boilerplate

PR #319, opened 2020-01-22
Event processing - Stream Processors

woksin:
Implemented the base of the StreamProcessor and created the interfaces and concepts it needs to talk with. 

The beginning of the foundation for processing streams of events.

There is just a minor thing missing from the StreamProcesser which has to do with retry timeout and keeping time with a clock or timer. 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1115616103797671/1158313806948768)


Specifications/Events.Processing/mocks.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.Logging;
+using Moq;
+
+namespace Dolittle.Runtime.Events
+{
+    public static class mocks
+    {
+        public static Mock<ILogger> a_logger()
```
einari: I would just use `Mock.Of<ILogger>()` in the place where it is using logging.

Source/Events.Processing/ProcessingResultValue.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Dolittle.Runtime.Events.Processing
+{
+    /// <summary>
+    /// Represents the state of the processing of an event.
+    /// </summary>
+    public enum ProcessingResultValue
```
einari: Value sounds a bit off for naming here.
Since its representing the state as it says in the API doc it could be called `ProcessingResultState`.

woksin: Agree. ProcessingState it is

Specifications/Events.Processing/mocks.cs
woksin: Cool. I removed it

PR #331, opened 2020-02-07
Make it work after testing with MongoDB event store

woksin:
* Needed to formalize some exceptions that can get thrown in event store.
* Stream processors now process events as fast as they can, but wait 1 sec for each poll when no more events.
* Make IEventStore interfaces async
* CancellationToken on all async methods
* Get specs working

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1160966669029431)

### DIFF
--- a/Source/Events/AggregateRoot.cs
+++ b/Source/Events/AggregateRoot.cs
@@ -183,7 +183,7 @@ namespace Dolittle.Events
             if (@event.AggregateRootVersion != Version) throw new AggregateRootVersionIsOutOfOrder(@event.AggregateRootVersion, Version);
         }
 
-        void ThrowIfStatefull()
+        void ThrowIfStateful()
         {
             if (!this.IsStateless()) throw new FastForwardNotAllowedForStatefulAggregateRoot(GetType());
         }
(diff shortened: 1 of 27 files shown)


Source/Events/AggregateRoot.cs
```diff
@@ -183,7 +183,7 @@ void ThrowIfAggreggateRootVersionIsOutOfOrder(CommittedAggregateEvent @event)
             if (@event.AggregateRootVersion != Version) throw new AggregateRootVersionIsOutOfOrder(@event.AggregateRootVersion, Version);
         }
 
-        void ThrowIfStateful()
+        void ThrowIfStatefull()
```
einari: It was correct before :) 

```suggestion
        void ThrowIfStateful()
```

Source/Events/AggregateRoot.cs
```diff
@@ -131,10 +131,10 @@ public virtual void ReApply(CommittedAggregateEvents events)
         /// <summary>
         /// Fast forward to the specified version of the <see cref="EventSource">EventSource</see>.
         /// </summary>
-        /// <param name="version">Version to fast foward to.</param>
+        /// <param name="version">Version to fast forward to.</param>
         public void FastForward(AggregateRootVersion version)
         {
-            ThrowIfStateful();
+            ThrowIfStatefull();
```
einari: It was correct before :)  Stateful not Statefull

```suggestion
            ThrowIfStateful();
```

PR #334, opened 2020-02-13
Specs for the new event store + bug fixes

woksin:
* Created specs for most of the functionallity of the event store
* Fixed bugs that surfaced through specs

Specs for the event store are utilzing Mongo2Go for real integration testing with mongo. Which proved itself since it surfaced some errors!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1161878374506982)


woksin:
Disclaimer! The tests using mongo2go is pretty slow. Need to figure that out

PR #336, opened 2020-02-15
Sdk hookup

einari:
This pull request does a couple of things:

- Simplifying levels of indirection. (Remote* removed)
- Introducing an abstract base filter
- Changing names to be more consistent with how we have other things - removing post fixes
- Provides service endpoints for event handlers and filters
- Started on formalizing type with partition filter and more specifically its definition and what should be persisted.

There are some dangling things and missing specs for a few things.
The dangling things have their own issues registered, while missing specs has been registered as technical debt that will be covered in the next month.

Primarily this PR is all about enabling the SDKs to be able to hookup.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1162122550910774)


Source/Events.Processing/FilterProcessor.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                Event = @event.ToProtobuf(),
+                ExecutionContext = _executionContextManager.Current.ToByteString()
+            };
+
+            _logger.Debug($"Filter event that occurred @ {@event.Occurred}");
+            FilterClientToRuntimeResponse result = null;
+            await _callDispatcher.Call(message, response => result = response).ConfigureAwait(false);
+            _logger.Debug($"Filter result : {result.IsIncluded}");
+            return new SucceededFilteringResult(true, partitionId);
```
woksin: I guess that it needs to write event to stream here based on if the result says that it is included in stream

einari: This is handled by the AbstractFilterProcessor that this one implements in its `Process()` method.

woksin: Oh, yeah of course :P

PR #363, opened 2020-03-25
Use CancellationToken where it should

woksin:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1168319876794588)


dolittle-unito:
➤ Einar Ingebrigtsen commented:

I pulled this in - couldn't see anything fundamentally flawed here and also we're doing the holistic code-review, hoping we're catching the big issues from that.

PR #366, opened 2020-04-03
Improve event horizon subscription + simplifications

woksin:
Based off the grpc contracts from this PR https://github.com/dolittle-runtime/Contracts/pull/13 
This PR attempts to simplify the event horizon subscription process, mostly on the consumer runtime side of things. It also simplifies across the board with classes related to the subscription services.

Uses Dolittle.Resilience for endlessly trying to connect/reconnect to producer runtimes

I discovered a problem when testing this. When killing the dotnet processes there are still some lingering processes related to the grpc connections that won't kill themselves. It may or may not be directly related to this new implementation. We need to look into it.

EDIT:
The commit https://github.com/dolittle-runtime/Runtime/pull/366/commits/ff1065e48b52c2cc84dc27386102457f4502b8fe fixes the problem mentioned above by properly disposing the consumer service server and properly dealing with cancellation tokens in the consumer client. I would though like a couple more eyes on it. @jakhog @einari. I've tested this implementation on with my testing scenario and it seems to give a quite stable reconnecting event horizon connection 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1169777311351456)


Source/Logging.Management/ILogManager.cs
```diff
@@ -12,14 +11,14 @@ namespace Dolittle.Runtime.Logging.Management
     public interface ILogManager
     {
         /// <summary>
-        /// Gets a <see cref="ObservableCollection{T}"/> of <see cref="JsonLogMessage">log messages</see>.
+        /// Gets a <see cref="ObservableCollection{T}"/> of <see cref="string">log messages</see>.
         /// </summary>
-        ObservableCollection<JsonLogMessage> Messages { get; }
+        ObservableCollection<string> Messages { get; }
```
woksin, line 16: We no longer have the JsonLogMessage construct in the fundamentals. This should perhaps be done in a better way, but can we roll with this for now? @einari @jakhog

jakhog:
We have not reviewed any file in Source/EventHorizon

PR #367, opened 2020-04-03
Runtime metrics

einari:
This adds metrics for event store. 
The following metrics, which are labeled for accuracy:

- CommittedEvents
- CommittedAggregateEvents
- UncommittedEvents
- UncommittedAggregateEvents
- FailedUncommittedEvents
- FailedUncommittedAggregateEvents

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1169873769642248)


jakhog:
This is what I recall from our discussion on these metrics:

1. Uncommitted + Comitted + Failed is more than we need. Every uncommitted event either becomes a comitted event or fails, so the latter two should be enough.

2. The comitted aggregate events, are really building upon the normal events. So I think it would make sense that when incrementing aggregate events, both counters are incremented. We still have all the information to get back to number of non-aggregate-events committed if we want to, but I think the most interesting number is going to be the total number of evens. It is nice if we try to keep the calculation of the most relevant numbers in the code, that makes the job in terms of analysing and putting alert rules on the runtime monitoring system a lot easier.

3. There is quite a lot of labels on the metrics just now. I think we agreed to reduce them a bit. Unfortunately I don't remember exactly which ones we decided to keep...

joelhoisko:
I changed the target branch into master, merged our current stuff into this branch and then did 3 changes based on Jakobs suggestions:

1. Removed the `UncommittedEvents` metrics.
2. Removed the word `Uncommitted` from the failed event counters/methods etc. Or is it important that we keep the `Uncommitted` wording in there?
3. Made incrementing `CommittedAggregateEvents` also increase `CommittedEvents`.

woksin:
Is this PR ready for review now, @joelhoisko ?

Source/Events.Store/MetricsProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MetricsProvider"/> class.
+        /// </summary>
+        /// <param name="metrics">basdasd.</param>
+        public MetricsProvider(IMetrics metrics)
+        {
+            _metrics = metrics;
+        }
+    }
+}
```
woksin, line 22: How is this class used?

joelhoisko: Is not, removed it 👍

Source/Events.Store/IMetrics.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        /// <param name="event">The <see cref="CommittedEvent"/> used for labelling.</param>
+        void IncrementCommittedEvents(CommittedEvent @event);
+
+        /// <summary>
+        /// Increment number of committed aggregate events and committed events based on an
+        /// event for labelling.
+        /// </summary>
+        /// <param name="event">The <see cref="CommittedAggregateEvent"/> used for labelling.</param>
+        void IncrementCommittedAggregateEvents(CommittedAggregateEvent @event);
```
woksin, line 56: The APIs on these should be similar IMO. Now the parameter for incrementing failed commits is plural while for non-failures its not. I suggest plural (Committed(Aggregate)Events, pretty sure we have a class for that?)

Source/Events.Store.MongoDB/EventStore.cs
```diff
@@ -75,6 +79,8 @@ public async Task<CommittedEvents> CommitEvents(UncommittedEvents events, Cancel
                                 cancel).ConfigureAwait(false);
                             committedEvents.Add(committedEvent);
                             eventLogSequenceNumber++;
+
+                            _metrics.IncrementCommittedEvents(committedEvent);
                         }
 
                         return new CommittedEvents(committedEvents);
```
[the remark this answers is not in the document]
woksin: And this is a transaction, so if one of the commit fails none of them goes through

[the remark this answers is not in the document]
joelhoisko: But it's in a for-loop and we do want to wait for the event to be committed fully until we increase the counter. If the `CommitEvent` method fails then we do not wan't to increment the counter, instead we catch the error later and increment the failed events counter.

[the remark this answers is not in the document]
woksin: Yes, and that's why putting it in the end will work. If it throws an exception it will not increase the committed counter, and if one of the commits in the transaction fails it will also not increment the committed counter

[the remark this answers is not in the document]
joelhoisko: Oh whoops, hadn't refreshed the page so didn't see your comment about the transaction. Was confused as I thought that it was possible for some of the events to succeed and some of them to fail.

Source/Events.Store/IMetrics.cs
```diff
@@ -45,14 +45,14 @@ public interface IMetrics
         /// <summary>
         /// Increment number of committed events based on an event for labelling.
         /// </summary>
-        /// <param name="event">The <see cref="CommittedEvent"/> used for labelling.</param>
-        void IncrementCommittedEvents(CommittedEvent @event);
+        /// <param name="event">The <see cref="CommittedEvents"/> used for labelling.</param>
+        void IncrementCommittedEvents(CommittedEvents @event);
```
woksin, line 49: plural 'events'
```suggestion
        /// <param name="events">The <see cref="CommittedEvents"/> used for labelling.</param>
        void IncrementCommittedEvents(CommittedEvents events);
```

Source/Events.Store/IMetrics.cs
```diff
@@ -45,14 +45,14 @@ public interface IMetrics
...
 
         /// <summary>
         /// Increment number of committed aggregate events and committed events based on an
         /// event for labelling.
         /// </summary>
-        /// <param name="event">The <see cref="CommittedAggregateEvent"/> used for labelling.</param>
-        void IncrementCommittedAggregateEvents(CommittedAggregateEvent @event);
+        /// <param name="event">The <see cref="CommittedAggregateEvents"/> used for labelling.</param>
+        void IncrementCommittedAggregateEvents(CommittedAggregateEvents @event);
```
woksin, line 56:
```suggestion
        /// <param name="events">The <see cref="CommittedAggregateEvents"/> used for labelling.</param>
        void IncrementCommittedAggregateEvents(CommittedAggregateEvents events);
```

joelhoisko:
Let's remember to update the first comment to adhere to our PR template when/if this gets pulled in.

woksin:
I hijacked this and made it updated for version 6

PR #368, opened 2020-04-06
Validation filter result

woksin:
Based on PR #366, pull that one in first.

Start on simplifying the filters. This is an ongoing process. More PRs will follow.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1170011480522006)


Source/EventHorizon/Producer/Filter/PublicFilterDefinition.cs
```diff
@@ -27,5 +27,8 @@ public PublicFilterDefinition(StreamId sourceStream, StreamId targetStream)
 
         /// <inheritdoc/>
         public StreamId TargetStream { get; }
+
+        /// <inheritdoc/>
+        public bool IsPersistable => false;
```
einari, line 32: You could have an interface `ICanBePersisted` instead - marker interface. Use this for checking if it should be persisted (`if (filterDefinition is ICanBePersisted) {}`)

einari:
suggestion: Use MongoDB Class Maps to represent the persistence: http://mongodb.github.io/mongo-csharp-driver/2.0/reference/bson/mapping/

Source/Events.Processing/Filters/FilterRegistry.cs
```diff
@@ -24,85 +20,57 @@ public class FilterRegistry : IFilterRegistry
...
+            if (!validationResult.Succeeded)
             {
-                _logger.Error(ex, $"No persisted filter repository for filter definition of type {typeof(TDefinition).FullName}");
+                var message = $"Filter '{filter.Definition.TargetStream}' failed validation: {validationResult.FailureReason}";
+                _logger.Debug(message);
+                throw new CouldNotRegisterFilter(filter.Definition, message);
             }
 
-            if (repository != null)
+            if (filter.Definition.IsPersistable)
```
einari, line 62: Use the `ICanBePersisted` interface to check for if it should be persisted

einari:
Suggest also looking into the Concept serializer: https://github.com/dolittle-extensions/ReadModels.MongoDB/blob/master/Source/ConceptSerializationProvider.cs

Source/EventHorizon/Producer/Filter/PublicFilterDefinition.cs
einari: Or...  `IPersistedFilterDefinition`

woksin:
Closing this. This is coming as a part of another PR.

PR #372, opened 2020-04-17
Change streampositionrange constructor to use "to" and "length" parameters

joelhoisko:
With the use of `(from, length)` style consturctor we don't need to have exceptions dealing with comparing the old `(from, to)`, like that `to` is always greater than `from` etc.

Simplified example of the new range:
`StreamPositionRange(0, 1)` would return the first element
`StreamPositionRange(5, 1)` would return the 5th element
`StreamPositionRange(50, 100)` would return elements 50-149

Do note that this is not how the range functions usually work in some common programming languages, like in [C#](https://docs.microsoft.com/en-us/dotnet/api/system.range.-ctor?view=netcore-3.1#System_Range__ctor_System_Index_System_Index_) or [Python](https://docs.python.org/3/library/stdtypes.html#typesseq-range) as they use the common `(from, to)` style.

I also relined some of the really long statements, I'm not quite sure what's our/C# practice for formatting multi-line stuff so feedback is welcome.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1171734720050260)


Source/Events.Processing/Filters/RemoteFilterValidator.cs
```diff
@@ -42,10 +42,25 @@ public RemoteFilterValidator(IFetchEventsFromStreams eventsFromStreams, IFetchEv
...
+                filter.Definition.TargetStream,
+                new StreamPositionRange(StreamPosition.Start, uint.MaxValue),
+                cancellationToken)
+                .ConfigureAwait(false);
+            var sourceStreamEvents = lastUnProcessedEventPosition == 0
+                ? Enumerable.Empty<StreamEvent>()
+                : await _eventsFromStreams.FetchRange(
+                    filter.Scope,
+                    filter.Definition.SourceStream,
+                    new StreamPositionRange(StreamPosition.Start, lastUnProcessedEventPosition - StreamPosition.Start),
```
woksin, line 61: Why lastUnProcessedEventPosition - StreamPosition.Start? StreamPosition.Start is alway == 0

woksin: Could just be lastUnProcessedEventPosition

Source/Events.Processing/Filters/TypeFilterWithEventSourcePartitionValidator.cs
```diff
@@ -53,10 +53,26 @@ public class TypeFilterWithEventSourcePartitionValidator : ICanValidateFilterFor
...
+                    filter.Definition.TargetStream,
+                    new StreamPositionRange(StreamPosition.Start, uint.MaxValue),
+                    cancellationToken)
+                .ConfigureAwait(false);
+            var sourceStreamEvents = lastUnProcessedEventPosition == 0
+                ? Enumerable.Empty<StreamEvent>()
+                : await _eventsFromStreams.FetchRange(
+                    filter.Scope,
+                    filter.Definition.SourceStream,
+                    new StreamPositionRange(StreamPosition.Start, lastUnProcessedEventPosition - StreamPosition.Start),
```
woksin, line 73: Same as above

Source/Events.Streams/StreamPositionRange.cs
```diff
@@ -27,11 +27,11 @@ public StreamPositionRange(StreamPosition from, StreamPosition to)
...
         /// <summary>
-        /// Gets the, inclusive, end of the range.
+        /// Gets the length of the wanted range.
         /// </summary>
-        public StreamPosition To { get; }
+        public ulong Length { get; }
 
         /// <inheritdoc/>
-        public override string ToString() => $"({From}, {To})";
+        public override string ToString() => $"({From}, {Length})";
```
woksin, line 35: The ToString could instead of being (from, length) be (from, to) where to = from + length. Then it would be the same as was before

woksin: Does that make sense?

PR #373, opened 2020-04-20
Move events.streams

petriw:
Two commits, first is moving files required for namespace change, and second is the namespace change itself.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1171956738936328)


woksin:
Nice. I merged it and resolved conflicts

PR #374, opened 2020-04-20
Add abstract Event and EventSequence classes to be used by Uncomitted/Committed/Aggregate-Events

joelhoisko:
This joins up some of the duplicate code by putting it into the abstract base classes instead. 

Also have to be explicit now about what `Event` you're using so that's why all the `MongoDB.Event.Event` declarations.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172012965612225)


Source/Events.Store/CommittedEventSequence.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Dolittle.Runtime.Events.Store
+{
+    /// <summary>
+    /// Sequence of committed events.
```
woksin, line 9: We usually for class start the documentation with 'Represents'. And we use the <see> tag for class references where fitting 
```suggestion
    /// Represents the basis for a sequence of committed <see cref="Event" >events</see>.
```

Source/Events.Store/Event.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.Artifacts;
+
+namespace Dolittle.Runtime.Events.Store
+{
+    /// <summary>
+    /// Abstract type for Committed/UncommittedEvent.
```
woksin, line 9: For abstract classes we usually say that it is a 'basis' for something.
```suggestion
    /// Represents the basis for an event.
```

Source/Events.Store/EventSequence.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using Dolittle.Collections;
+
+namespace Dolittle.Runtime.Events.Store
+{
+    /// <summary>
+    /// Sequence of events.
```
woksin, line 11: Same as above
```suggestion
    /// Represents the basis for a sequence of <see cref="Event" >events</see>.
```

PR #377, opened 2020-04-22
Massive pr

woksin:
Lots of files being edited here. This is a long-lived branch with lots of different stuff. Though most of it is removal of obsolete files and fixing specs.

Ready:
* Fixed most of the specs so that they are building and passing
* Removed some obsolete specs
* Marked some specs as [Ignored] as they need some more effort
* Removes *.Management and TimeSeries projects
* Dolittle.Applications => Dolittle.ApplicationModel
* Use new Contracts. Without the 'external alias' thing
* Remove 'Cause'
* Remove 'RetryProcessingState' from IEventProcessor
* Remove 'ProcessorFailureType'
* Add new Process-method to IEventProcessor with 'failureReason' and 'retryCount' that is used only for stream processors that are catching up
* Move PublicFilter to Events.Processing
* Separate StreamProcessor registration from the starting of a StreamProcessor. The one that registers the StreamProcessor is responsible for the lifecycle of it. Starting it and Disposing it when it is supposed to be disposed.
* The EventHandlers and Filters services will attempt to register all StreamProcessors before starting them. If the registration of a StreamProcessor is failed for any of the tenants then the service will write a failed response and close the connection
* Removed FilterRegistry as it is not needed. There is no point in keeping an in memory registry of the filter definitions. The filter processors are StreamProcessors which are registered.
* Removed Bindings for 'ICanValidateFilter'. The filter validators are discovered and registered in the IFilterValidators 
* Create 'FilterValidationResult' which is returned by the FilterValidator.Validate
* Get rid of 'ICanRemovePersistedFilterDefinition'
* Add 'IFilterDefinitionRepository' which can persist and retrieve 'IPersistableFilter' filters
* 'StreamProcessor's Unregister themselves from the 'StreamProcessors' registry when cancellation is requested. It is also defined as an IDisposable now. Unregistering a stream processor multiple times should not throw any exception or do anything unexpected.
* 'StreamProcessors' returns a 'StreamProcessorRegistrationResult' that, among other things, says whether the stream processor has already been registered or not. So the caller can for itself decide what to do.
* Have a general filter definition repository in Mongo event store
* Remove IDisposable interface from IEventStore
* 'CommittedEvent's has ExecutionContext as a property
* Events.Store.MongoDB ExecutionContext: Added Version, Environment.
* Adhere to changes in Fundamentals

Half-way:
* Start on using the new formated log functions
* Start on formalizing FailureId's for known errors on services.
* Maybe find some common functionality that can be shared between the *Service classes.
* When a 'StreamProcessor' starts, the task is stored and returned en future calls to Start(). The method Start does not necessarily reflect this. We need to decide what we want the semantics to be. Either consecutive calls on Start throws an exception or it needs to do something else, for example returning the original task as I do already.
* Removed unused Exceptions
* Removed most of the code in Heads. More of it may also be obsolete

Not ready:
* EventHorizonEvent's ExecutionContext is not set right. The Tenant is set to be the consumer tenant, it should be the producer tenant.
* Need to implement the PublicFilterValidator validate function 
* The IFilter, IEventHandlers stuff may perhaps not be necessary as they don't really do that much. They only register the stream processors and returns a result.
* The filter validators are still not "correct"
* Use BsonClassMaps and Serializers in Events.Store.MongoDB 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172415325250689)


Source/Events.Processing/Filters/FailedFilterRegistrationReason.cs
```diff
@@ -0,0 +1,42 @@
...
+                string.Empty
+                : $"{(string.IsNullOrEmpty(value) ? string.Empty : "\n")}Stream Processor for filter with Stream Processor Id: '{filterProcessorRegistrationResult.StreamProcessor.Identifier}' has already been registered";
+
+            value += filterValidationResult.Succeeded ?
+                string.Empty
+                : $"{(string.IsNullOrEmpty(value) ? string.Empty : "\n")}Filter validation failed. {filterValidationResult.FailureReason}";
+
+            return new FailedFilterRegistrationReason { Value = value };
+        }
+    }
```
woksin, line 41: Same with the other one, I don't like it

Source/Events.Processing/Filters/FailedFilterValidationReason.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Dolittle.Runtime.Events.Processing.Filters
+{
+    /// <summary>
+    /// Represents the reason for why the validation of a <see cref="IFilterDefinition" /> failed.
+    /// </summary>
+    public class FailedFilterValidationReason : Concepts.ConceptAs<string>
```
woksin, line 9: Use 'using Dolittle.Concepts'

Source/Events.Processing/Filters/FilterValidators.cs
```diff
@@ -2,50 +2,99 @@
...
+        /// <param name="typeFinder">The <see cref="ITypeFinder" />.</param>
         /// <param name="container">The <see cref="IContainer" />.</param>
         /// <param name="logger">The <see cref="ILogger" />.</param>
-        public FilterValidators(IContainer container, ILogger logger)
+        public FilterValidators(ITypeFinder typeFinder, IContainer container, ILogger logger)
         {
+            _typeFinder = typeFinder;
             _container = container;
             _logger = logger;
+            PopulateFilterValidatorMap();
```
woksin, line 39: Instead of doing this we could have an ICanProvideBindings which binds them up using the TypeFinder.

woksin:
Pulling this abomination in, in agreement with @jakhog

PR #378, opened 2020-04-24
Split up Filters services

woksin:
Based off the "MassivePR" #377 branch. Pull that one inn first

We now differentiate between partitioned and unpartitioned filters. This PR makes the Runtime comply to the new definitions in the Contracts that will come with this PR https://github.com/dolittle-runtime/Contracts/pull/16

Adds:
* Partitioned FilterProcessor and FilterService
* UnPartitioned FilterProcessor and FilterService

Changes:
* Fixes Specs for Events.Processing to comply what's added and changed here
* IFilterResult has now a 'IsPartitioned' property
* SuccessfulFiltering has now two constructors; one with 'PartitionId' and one without. We could differentiate between IPartitionedFilterResult and IUnPartitionedFilterResult, but that depends a bit on how we want to further formalize the partitioned streams stuff; if we want to have I(Un)PartionedEventProcessor and such

Changes that has to be made in the future:
* When registering filters we now need to know if that filter has been registered before and whether it was registered as a partitioned or unpartitioned filter. That might mean that we have to store all filter definitions and the filter definitions themselves need a property to signify whether it is defines a partitioned or unpartitioned filter.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172713660945302)


woksin:
I've change the contract names a bit on https://github.com/dolittle-runtime/Contracts/pull/16

I'll update this to match

woksin:
The last commit changes the filter services to comply to the new change requested in the contracts so that everything builds. This is not how it is supposed to look! My [next PR](
https://github.com/dolittle-runtime/Runtime/pull/382) greatly enhances this on the service side, so I will make it look nice there.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+            try
+            {
+                if (!PartitionedTryStartStreamProcessors(registrationResults, out var failure))
+                {
+                    _logger.Warning(failure.Reason);
+                    await PartitionedWriteFailedRegistrationResponse(dispatcher, failure, context.CancellationToken).ConfigureAwait(false);
+                    return;
+                }
+
+                await dispatcher.Accept(new FilterRegistrationResponse(), context.CancellationToken).ConfigureAwait(false);
```
jakhog, line 186: It seems here like you start the StreamProcessors before the ReverseCall is accepted - which could then potentially start sending calls before the response is sent to the client - which is not allowed. Seems like there needs to be a two step registration here, where the Filters are created and registered (so they can't be started twice), and if that is succesfull, the ReverseCall should be accepted, and then the processing can commence.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+
+            var arguments = dispatcher.Arguments;
+            var filterId = arguments.FilterId.To<StreamId>();
+            if (filterId.IsNonWriteable)
+            {
+                _logger.Warning("Received public filter registration request with Filter Id: '{filterId}' which is an invalid stream id", filterId);
+                var failure = new Failure(FiltersFailures.CannotRegisterFilterOnNonWriteableStream, $"Received public filter registration request with Filter Id: '{filterId}' which is an invalid stream id");
+                await PublicWriteFailedRegistrationResponse(dispatcher, failure, context.CancellationToken).ConfigureAwait(false);
+                return;
+            }
```
jakhog, line 231: This logic is duplicated multiple times - just with different response types. So that could be pulled out to a generic method. But that might be one of those things that you were thinking of improving later either way.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+            try
+            {
+                if (!PublicTryStartStreamProcessors(registrationResults, out var failure))
+                {
+                    _logger.Warning(failure.Reason);
+                    await PublicWriteFailedRegistrationResponse(dispatcher, failure, context.CancellationToken).ConfigureAwait(false);
+                    return;
+                }
+
+                await dispatcher.Accept(new FilterRegistrationResponse(), context.CancellationToken).ConfigureAwait(false);
```
jakhog, line 249: Same here with Call before Accept.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+            }
+            finally
+            {
+                registrationResults
+                    .SelectMany(tenantAndResult => new[] { tenantAndResult.Item2.FilterStreamProcessor })
+                    .ForEach(_ => _.Dispose());
+            }
+        }
+
         bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<RemoteFilterDefinition>)> registrationResults, out Failure failure)
```
jakhog, line 259: I'm not very fond of the `.Item1` etc. that end up in the code when unnamed tuples are used. If we replace this with something like this, we can refer to the items as normal properties:
```suggestion
        bool TryStartStreamProcessors(IEnumerable<(TenantId Tenant, FilterRegistrationResult<RemoteFilterDefinition> Result)> registrationResults, out Failure failure)
```

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
             IReverseCallDispatcher<FiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, FiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, FilterResponse> dispatcher,
             Failure failure,
             CancellationToken cancellationToken) => dispatcher.Reject(new FilterRegistrationResponse { Failure = failure }, cancellationToken);
+
+        bool PartitionedTryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<RemoteFilterDefinition>)> registrationResults, out Failure failure)
```
jakhog, line 311: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+
+            foreach ((var tenant, var result) in registrationResults)
+            {
+                _ = result.FilterStreamProcessor.Start();
+            }
+
+            return true;
+        }
+
+        async Task<IEnumerable<(TenantId, FilterRegistrationResult<RemoteFilterDefinition>)>> PartitionedRegisterStreamProcessorsForAllTenants(
```
jakhog, line 333: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+            ScopeId scope,
+            StreamId sourceStream,
+            StreamId targetStream,
+            IReverseCallDispatcher<PartitionedFiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, PartitionedFiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, PartitionedFilterResponse> dispatcher,
+            CancellationToken cancellationToken)
+        {
+            var registrationResults = new List<(TenantId, FilterRegistrationResult<RemoteFilterDefinition>)>();
```
jakhog, line 340: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+
+            return registrationResults.AsEnumerable();
+        }
+
+        Task PartitionedWriteFailedRegistrationResponse(
+            IReverseCallDispatcher<PartitionedFiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, PartitionedFiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, PartitionedFilterResponse> dispatcher,
+            Failure failure,
+            CancellationToken cancellationToken) => dispatcher.Reject(new FilterRegistrationResponse { Failure = failure }, cancellationToken);
+
+        bool PublicTryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<PublicFilterDefinition>)> registrationResults, out Failure failure)
```
jakhog, line 363: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+
+            foreach ((var tenant, var result) in registrationResults)
+            {
+                _ = result.FilterStreamProcessor.Start();
+            }
+
+            return true;
+        }
+
+        async Task<IEnumerable<(TenantId, FilterRegistrationResult<PublicFilterDefinition>)>> PublicRegisterStreamProcessorsForAllTenants(
```
jakhog, line 385: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+            StreamId sourceStream,
+            StreamId targetStream,
+            IReverseCallDispatcher<PublicFiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, PublicFiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, PartitionedFilterResponse> dispatcher,
+            CancellationToken cancellationToken)
+        {
+            var registrationResults = new List<(TenantId, FilterRegistrationResult<PublicFilterDefinition>)>();
```
jakhog, line 391: Make tuple named.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+                    new RemoteFilterDefinition(sourceStream, targetStream),
+                    dispatcher,
+                    _getEventsToStreamsWriter(),
+                    _logger);
+                var filters = _getFilters();
+                var registrationResult = await filters.Register(filterProcessor, cancellationToken).ConfigureAwait(false);
+                registrationResults.Add((tenant, registrationResult));
+            }
+
+            return registrationResults.AsEnumerable();
```
jakhog, line 355: The `List<..>` implements `IEnumerable` so no need to convert it.
```suggestion
            return registrationResults;
```

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+                    new PublicFilterDefinition(sourceStream, targetStream),
+                    dispatcher,
+                    _getEventsToPublicStreamsWriter(),
+                    _logger);
+                var filters = _getFilters();
+                var registrationResult = await filters.Register(filterProcessor, cancellationToken).ConfigureAwait(false);
+                registrationResults.Add((tenant, registrationResult));
+            }
+
+            return registrationResults.AsEnumerable();
```
jakhog, line 405: No need to use `.AsEnumerable()`.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+                    await PartitionedWriteFailedRegistrationResponse(dispatcher, failure, context.CancellationToken).ConfigureAwait(false);
+                    return;
+                }
+
+                await dispatcher.Accept(new FilterRegistrationResponse(), context.CancellationToken).ConfigureAwait(false);
+            }
+            finally
+            {
+                registrationResults
+                    .SelectMany(tenantAndResult => new[] { tenantAndResult.Item2.FilterStreamProcessor })
```
jakhog, line 191: Use named property here after changing to named tuples.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -123,6 +128,134 @@ public class FiltersService : FiltersBase
...
+                    await PublicWriteFailedRegistrationResponse(dispatcher, failure, context.CancellationToken).ConfigureAwait(false);
+                    return;
+                }
+
+                await dispatcher.Accept(new FilterRegistrationResponse(), context.CancellationToken).ConfigureAwait(false);
+            }
+            finally
+            {
+                registrationResults
+                    .SelectMany(tenantAndResult => new[] { tenantAndResult.Item2.FilterStreamProcessor })
```
jakhog, line 254: Use named property here.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
             IReverseCallDispatcher<FiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, FiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, FilterResponse> dispatcher,
             Failure failure,
             CancellationToken cancellationToken) => dispatcher.Reject(new FilterRegistrationResponse { Failure = failure }, cancellationToken);
+
+        bool PartitionedTryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<RemoteFilterDefinition>)> registrationResults, out Failure failure)
+        {
+            failure = null;
+            var failedRegistrationReasons = registrationResults
+                                                .Where(tenantAndResult => !tenantAndResult.Item2.Succeeded)
+                                                .Select(tenantAndResult => $"For tenant '{tenantAndResult.Item1}':\n\t{string.Join("\n\t", tenantAndResult.Item2.FailureReason.Value.Split("\n"))}");
```
jakhog, line 316: Use named property here.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+            IReverseCallDispatcher<PartitionedFiltersClientToRuntimeMessage, FilterRuntimeToClientMessage, PartitionedFiltersRegistrationRequest, FilterRegistrationResponse, FilterEventRequest, PartitionedFilterResponse> dispatcher,
+            Failure failure,
+            CancellationToken cancellationToken) => dispatcher.Reject(new FilterRegistrationResponse { Failure = failure }, cancellationToken);
+
+        bool PublicTryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<PublicFilterDefinition>)> registrationResults, out Failure failure)
+        {
+            failure = null;
+            var failedRegistrationReasons = registrationResults
+                                                .Where(tenantAndResult => !tenantAndResult.Item2.Succeeded)
+                                                .Select(tenantAndResult => $"For tenant '{tenantAndResult.Item1}':\n\t{string.Join("\n\t", tenantAndResult.Item2.FailureReason.Value.Split("\n"))}");
```
jakhog, line 368: Use named property here.

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+            {
+                _executionContextManager.CurrentFor(tenant);
+                var filterProcessor = new Partitioned.FilterProcessor(
+                    scope,
+                    new RemoteFilterDefinition(sourceStream, targetStream),
+                    dispatcher,
+                    _getEventsToStreamsWriter(),
+                    _logger);
+                var filters = _getFilters();
+                var registrationResult = await filters.Register(filterProcessor, cancellationToken).ConfigureAwait(false);
```
jakhog, line 351: It looks like the filters are registered one time per tenant? That seems unnecessary, since the same filter will be registered for all tenants, or none at all. So maybe the `IFilters` should be a singleton, and delegate the actual state things to other components...

Source/Events.Processing/Filters/FiltersService.cs
```diff
@@ -174,5 +307,107 @@ bool TryStartStreamProcessors(IEnumerable<(TenantId, FilterRegistrationResult<Re
...
+            {
+                var failureMessage = $"Failed to register filter:\n\t";
+                failureMessage += string.Join("\n\t", failedRegistrationReasons);
+                failure = new Failure(FiltersFailures.FailedToRegisterFilter, failureMessage);
+                return false;
+            }
+
+            foreach ((var tenant, var result) in registrationResults)
+            {
+                _ = result.FilterStreamProcessor.Start();
```
jakhog, line 327: What happens with this `Task`? It seems like the lifetime of that thread of execution is handled somewhere else, so I'm not sure it makes sense to return a `Task` on that method?

Source/Events.Processing/Filters/FiltersService.cs
woksin: I'll register a task to ensure that the dispatcher is ready to receive calls before the stream prcessors is started

Source/Events.Processing/Filters/FiltersService.cs
woksin: Resolved by later PR

woksin:
Registered tasks from the comments

PR #379, opened 2020-04-24
Move StreamProsessorId into StreamProcessorState and add compound indexes for it

joelhoisko:
Moves the EventStores `StreamProcessorId` into `StreamProcessorState`. These id's are then used to create a unique compound index in `EventStoreConnection`. Also fixed the necessary filters for it in `StreamProcessorStateRepository`.

I had to add a default `StreamProcessorState.Id` property so that MongoDB would stop crying about 
```
Element '_id' does not match any field or property of class StreamProcessorState...
```
in reality we'd want to have some sort of "compound id" consisting from the 3 id properties but I don't know how to do that.

Also renamed lots of methods as they only create indexes, not collections.

Tested with the triple trouble.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172725482949905)


Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -40,7 +40,7 @@ public EventStoreConnection(DatabaseConnection connection, ILogger logger)
             StreamProcessorStates = connection.Database.GetCollection<StreamProcessorState>(Constants.StreamProcessorStateCollection);
             TypePartitionFilterDefinitions = connection.Database.GetCollection<TypePartitionFilterDefinition>(Constants.TypePartitionFilterDefinitionCollection);
 
-            CreateCollectionsAndIndexes();
+            CreateIndexes();
```
woksin, line 43: When creating the indexes as we do here, Mongo implicitly creates the collection. Because we're using a mongo replica set we need to explicitly create the collections before we do any other operations on it (as far as I understand), so when the event store connection is established we want to create indexes and thus also implicitly create the collections, hence the CreateCollectionsAndIndexes name :)

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -87,7 +87,7 @@ public EventStoreConnection(DatabaseConnection connection, ILogger logger)
         public async Task<IMongoCollection<Events.StreamEvent>> GetStreamCollection(StreamId stream, CancellationToken cancellationToken)
         {
             var collection = _connection.Database.GetCollection<Events.StreamEvent>(Constants.CollectionNameForStream(stream));
-            await CreateCollectionsAndIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
+            await CreateIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
```
woksin, line 90: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -101,7 +101,7 @@ public async Task<IMongoCollection<Events.StreamEvent>> GetStreamCollection(Stre
         public async Task<IMongoCollection<Events.StreamEvent>> GetScopedStreamCollection(ScopeId scope, StreamId stream, CancellationToken cancellationToken)
         {
             var collection = _connection.Database.GetCollection<Events.StreamEvent>(Constants.CollectionNameForScopedStream(scope, stream));
-            await CreateCollectionsAndIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
+            await CreateIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
```
woksin, line 104: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -128,7 +128,7 @@ public async Task<IMongoCollection<Events.StreamEvent>> GetScopedStreamCollectio
             CancellationToken cancellationToken)
         {
             var collection = _connection.Database.GetCollection<StreamProcessorState>(Constants.CollectionNameForScopedStreamProcessorStates(scope));
-            await CreateCollectionsAndIndexesForStreamProcessorStatesAsync(collection, cancellationToken).ConfigureAwait(false);
+            await CreateIndexesForStreamProcessorStatesAsync(collection, cancellationToken).ConfigureAwait(false);
```
woksin, line 131: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -150,7 +150,7 @@ public async Task<IMongoCollection<Events.StreamEvent>> GetScopedStreamCollectio
         public async Task<IMongoCollection<MongoDB.Events.Event>> GetScopedEventLog(ScopeId scope, CancellationToken cancellationToken)
         {
             var collection = _connection.Database.GetCollection<MongoDB.Events.Event>(Constants.CollectionNameForScopedEventLog(scope));
-            await CreateCollectionsAndIndexesForEventLogAsync(collection, cancellationToken).ConfigureAwait(false);
+            await CreateIndexesForEventLogAsync(collection, cancellationToken).ConfigureAwait(false);
```
woksin, line 153: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -163,19 +163,19 @@ public async Task<IMongoCollection<MongoDB.Events.Event>> GetScopedEventLog(Scop
         public async Task<IMongoCollection<Events.StreamEvent>> GetPublicStreamCollection(StreamId stream, CancellationToken cancellationToken)
         {
             var collection = _connection.Database.GetCollection<Events.StreamEvent>(Constants.CollectionNameForPublicStream(stream));
-            await CreateCollectionsAndIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
+            await CreateIndexesForStreamEventsAsync(collection, cancellationToken).ConfigureAwait(false);
```
woksin, line 166: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -163,19 +163,19 @@ public async Task<IMongoCollection<MongoDB.Events.Event>> GetScopedEventLog(Scop
...
-            CreateCollectionsAndIndexesForStreamProcessorStates();
-            CreateCollectionsAndIndexesForTypePartitionFilterDefinitions();
+            CreateIndexesForEventLog();
+            CreateIndexesForAggregates();
+            CreateIndexesForStreamProcessorStates();
+            CreateIndexesForTypePartitionFilterDefinitions();
         }
 
-        void CreateCollectionsAndIndexesForEventLog()
+        void CreateIndexesForEventLog()
```
woksin, line 178: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -187,7 +187,7 @@ void CreateCollectionsAndIndexesForEventLog()
                     .Ascending(_ => _.Aggregate.TypeId)));
         }
 
-        void CreateCollectionsAndIndexesForAggregates()
+        void CreateIndexesForAggregates()
```
woksin, line 190: See above comment

Source/Events.Store.MongoDB/EventStoreConnection.cs
```diff
@@ -196,22 +196,27 @@ void CreateCollectionsAndIndexesForAggregates()
...
+        /// </summary>
+        void CreateIndexesForStreamProcessorStates()
         {
             StreamProcessorStates.Indexes.CreateOne(
                 new CreateIndexModel<StreamProcessorState>(
                 Builders<StreamProcessorState>.IndexKeys
-                    .Ascending(_ => _.Id)));
+                    .Ascending(_ => _.ScopeId)
+                    .Ascending(_ => _.EventProcessorId)
+                    .Ascending(_ => _.SourceStreamId)));
```
woksin, line 209: Cannot this also have the  new CreateIndexOptions { Unique = true }) ?

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorState.cs
```diff
@@ -17,21 +18,40 @@ public class StreamProcessorState
...
             Position = position;
             FailingPartitions = failingPartitions;
         }
 
         /// <summary>
-        /// Gets or sets the <see cref="StreamProcessorId" />.
+        /// Gets or sets the  MongoDB _id. This is used so that the class would have a valid '_id' field in mongo.
+        /// The classes 'true' id is comporomised from the combinaton of ScopeId, EventProcessorId and SourceStreamId.
+        /// </summary>
+        public ObjectId Id { get; set; }
```
woksin, line 39: I think that this is OK for now, but as you said there may be another solution. I'm not familiar enough with MongoDB to know

Source/Events.Store.MongoDB/EventStoreConnection.cs
woksin: Yes we could do that

PR #380, opened 2020-04-24
Persist partitioned filters

woksin:
Based off PR #378. Pull that in first.

All filters definitions are now either partitioned or not. Remote filters should then be persisted and validators should check whether the new filter definition and the persisted filter definition matches in terms of whether they are partitioned or not.

Adds:
* Add 'Partitioned' property on IFilterDefinition

Changes:
* Makes RemoteFilterDefinition persistable
* Partitioned and Unpartitioned filters persist their filter definitions

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172726683680134)


Source/Events.Processing/Filters/RemoteFilterDefinition.cs
```diff
@@ -8,23 +8,28 @@ namespace Dolittle.Runtime.Events.Processing.Filters
     /// <summary>
     /// Represents a <see cref="IFilterDefinition" /> for a remote filter.
     /// </summary>
-    public class RemoteFilterDefinition : IFilterDefinition
+    public class RemoteFilterDefinition : IPersistableFilterDefinition
```
jakhog, line 11: It looks a bit strange that in the core Runtime code, there is an `IFilterDefinition`, and then two implementations `RemoteFilterDefinition` and `TypedFilterWithEventSourcePartitionDefinition` - while in the MongoDB storage there are just `FilterDefinition` and `TypePartitionFilterDefinition` that correspond to the `RemoteFilterDefinition` and `TypedFilterWithEventSourcePartitionDefinition` respectively.

It seems like these two should represent eachother 1-to-1 ?

Source/Events.Store.MongoDB/Processing/Filters/FilterDefinitionRepository.cs
```diff
@@ -54,7 +54,7 @@ public async Task PersistFilter(IPersistableFilterDefinition filterDefinition, C
                 var document = filterDefinition switch
                     {
                         TypeFilterWithEventSourcePartitionDefinition definition => definition.ToStoreRepresentation(),
-                        _ => new FilterDefinition(filterDefinition.TargetStream, filterDefinition.SourceStream)
+                        _ => new FilterDefinition(filterDefinition.TargetStream, filterDefinition.SourceStream, filterDefinition.Partitioned)
```
jakhog, line 57: See previous comment - the logic here that the default here is `FilterDefinition` doesn't quite match up with that the _default_ in the core is `IPersistableFilterDefinition`.

Source/Events.Processing/Filters/RemoteFilterDefinition.cs
woksin: Yes, it is strange. Registering an issue for it.

PR #381, opened 2020-04-24
Action performer on tenants

woksin:
Creates two classes and interfaces which can perform actions on all currently registered Tenants. It will set the execution context for each tenant and then perform the action.

We could also extend those interfaces to be able to register actions which can be performed on all currently registered Tenants and all Tenants that will be added in the future. Could event register OnRemoved actions for when Tenants are removed from the collection for example...

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172731908767803)


Source/Tenancy/IPerformActionOnAllTenants.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using Dolittle.Execution;
+using Dolittle.Tenancy;
+
+namespace Dolittle.Runtime.Tenancy
+{
+    /// <summary>
+    /// Defines a system that perform an action for <see cref="ITenants.All" /> in an <see cref="ExecutionContext" /> with the tenant set for each Tenant.
+    /// </summary>
+    public interface IPerformActionOnAllTenants
```
einari, line 13: Suggest renaming either the class or the interface to be following the `IFoo` -> `Foo` convention, since this is not used for discovery and typically how we'd name it if there are multiple of these. 
That way you can get rid of the explicit binding for this and the convention will catch it.

Source/Tenancy/IPerformAsynchronousActionOnAllTenants.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+using Dolittle.Execution;
+using Dolittle.Tenancy;
+
+namespace Dolittle.Runtime.Tenancy
+{
+    /// <summary>
+    /// Defines a system that performs an asynchronous action for <see cref="ITenants.All" /> in an <see cref="ExecutionContext" /> with the tenant set for each Tenant.
+    /// </summary>
+    public interface IPerformAsynchronousActionOnAllTenants
```
einari, line 14: Suggestion; would it be an idea to actually just merge this with the other interface - just keep one and then differentiate on having `Perform` and `PerformAsync`.  The `PerformInParallel` would just keep its name.

Source/Tenancy/IPerformActionOnAllTenants.cs
einari: Interface could be something like : `ITenantsActionPerformer` implementation `TenantsActionPerformer`

woksin: Yeah, that sounds reasonable

Source/Tenancy/IPerformAsynchronousActionOnAllTenants.cs
woksin: Yeah could do that

woksin: Fixed by https://github.com/dolittle-runtime/Runtime/pull/381/commits/efba7f4fccbaf107e0cd31ac1eba6948f06e5c0b

Source/Tenancy/IPerformActionOnAllTenants.cs
woksin: Fixed by https://github.com/dolittle-runtime/Runtime/pull/381/commits/efba7f4fccbaf107e0cd31ac1eba6948f06e5c0b

PR #383, opened 2020-04-26
Partitioned and unpartitioned stream processo

woksin:
Based on the previous PR #382. Pull that in first

This PR differentiates between partitioned and unpartitioned stream processors

I purposefully omitted fixing the specs since I recon that there will be quite a few files that has to be fixed. I can make a new branch and maybe another PR if we don't want it into this.

* IStreamProcessorStates: A system that knows about persisted stream processor states. There are also two other interfaces that comes from this, one for partitioned- and one for unpartitioned stream processor states.
* Partitioned and Unpartitioned stream processor. The behaviour should not have changed

Changes that needs to be done later:
* Make specs build and pass

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1172896783889971)


Source/Events.Processing/Filters/RemoteFilterValidator.cs
```diff
@@ -22,28 +22,32 @@ public class RemoteFilterValidator : ICanValidateFilterFor<RemoteFilterDefinitio
...
-            _streamProcessorStateRepository = streamProcessorStateRepository;
+            _streamProcessorStateRepository = streamProcessorStates;
             _logger = logger;
         }
 
         /// <inheritdoc/>
         public async Task<FilterValidationResult> Validate(IFilterProcessor<RemoteFilterDefinition> filter, CancellationToken cancellationToken)
         {
-            var streamProcessorState = await _streamProcessorStateRepository.GetOrAddNew(
+            var streamProcessorState = await _streamProcessorStateRepository.GetFor(
```
woksin, line 50: This will be null if there is no stream processor state persisted

Source/Events.Processing/Filters/TypeFilterWithEventSourcePartitionValidator.cs
```diff
@@ -50,8 +50,8 @@ public class TypeFilterWithEventSourcePartitionValidator : ICanValidateFilterFor
 
         async Task<FilterValidationResult> ValidateBasedOffReFilteredStream(IFilterProcessor<TypeFilterWithEventSourcePartitionDefinition> filter, CancellationToken cancellationToken)
         {
-            var streamProcessorState = await _streamProcessorStateRepository
-                .GetOrAddNew(
+            var streamProcessorState = await _streamProcessorStates
+                .GetFor(
```
woksin, line 54: This will be null if there is no stream processor state persisted

woksin:
This will be the last of the monsters! At the end there I just commented out the code in the stream processor state files in Events.Store.MongoDB. Because @joelhoisko will fix that :)

woksin:
Dragging this in so that we're already running with this code and testing. We should together have a review later as well

PR #387, opened 2020-05-05
Use BsonType.Document for DateTimeOffsets

joelhoisko:
We can't do nice queries for the `DateTimeOffsets` in mongo when they're just strings. Using `BsonType.Document` saves a UTCfied DateTime, the ticks and offset (in minutes) to mongo so that it is nicer to look at and query for.

Do note that the  converted  `DateTime` loses some of the info as it's converted to UTC. This field should only be used for visual purposes and Mongos [DatetimeOffsetDeserializer](https://github.com/mongodb/mongo-csharp-driver/blob/21433508678992815f4d691dfa820263cea012e9/src/MongoDB.Bson/Serialization/Serializers/DateTimeOffsetSerializer.cs#L122) outright ignores it and deserializes the value from the ticks and offset.

I also changed the `DateTimeOffset` in `FailingPartitionState` to use `BsonType.Document`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1174244409722311)


Source/Events.Store.MongoDB/Events/EventMetadata.cs
```diff
@@ -32,8 +32,11 @@ public EventMetadata(DateTimeOffset occurred, Guid eventSource, Guid typeId, uin
 
         /// <summary>
         /// Gets or sets the <see cref="DateTimeOffset"/> of when the event was committed to the event store.
+        /// BsonType.Document saves a UTC DateTime, ticks and an offset(in minutes) to the document. This way we can
+        /// query for the DateTime from the database and it looks nicer than the string representation.
+        /// https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Serializers/DateTimeOffsetSerializer.cs#L158 .
```
woksin, line 37: This can be inside a <remarks></remarks> tags

Source/Events.Store.MongoDB/Processing/Streams/FailingPartitionState.cs
```diff
@@ -35,8 +35,11 @@ public FailingPartitionState(ulong position, DateTimeOffset retryTime, string re
 
         /// <summary>
         /// Gets or sets the retry time.
+        /// BsonType.Document saves a UTC DateTime, ticks and an offset(in minutes) to the document. This way we can
+        /// query for the DateTime from the database and it looks nicer than the string representation.
+        /// https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Serializers/DateTimeOffsetSerializer.cs#L158 .
```
woksin, line 40: Same as above

Source/Events.Store.MongoDB/Events/EventMetadata.cs
joelhoisko: added

Source/Events.Store.MongoDB/Processing/Streams/FailingPartitionState.cs
joelhoisko: added

PR #391, opened 2020-05-06
Stream processor timestamping

petriw:
This PR adds a DateTimeOffset to StreamProcessorState for the last successful process of a Steam, and a DateTimeOffset to FailingPartitionState for when it last failed.

We were not sure if the timestamp in FailingPartitionState should be the timestamp of when it last failed, or when it failed for the first time, or both. We went for Last Failed.

We could not test it yet since the SDK is still broken.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1174495017724461)


Source/Events.Processing/Streams/StreamProcessorState.cs
```diff
@@ -23,6 +23,22 @@ public StreamProcessorState(StreamPosition streamPosition)
             RetryTime = DateTimeOffset.UtcNow;
             FailureReason = string.Empty;
             ProcessingAttempts = 0;
```
woksin, line 25: Same comment as above with the constructor that does not make sense to expose publically since it should only be used once. Make it either private or have private setters on the properties

Source/Events.Processing/Streams/Partitioned/StreamProcessorState.cs
```diff
@@ -41,5 +56,10 @@ public StreamProcessorState(StreamPosition streamPosition, IDictionary<Partition
         /// Gets or sets the <see cref="IDictionary{PartitionId, FailingPartitionState}">states of the failing partitions</see>.
         /// </summary>
         public IDictionary<PartitionId, FailingPartitionState> FailingPartitions { get; set; }
+
+        /// <summary>
+        /// Gets the timestamp when the StreamProcessor has processed the stream.
+        /// </summary>
+        public DateTimeOffset LastSuccessfullyProcessed { get; }
```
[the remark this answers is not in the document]
petriw: Yes I can remove them :)

Source/Events.Processing/Streams/Partitioned/FailingPartitions.cs
```diff
@@ -44,12 +44,12 @@ public class FailingPartitions : IFailingPartitions
...
         public async Task<IStreamProcessorState> AddFailingPartitionFor(IStreamProcessorId streamProcessorId, StreamProcessorState oldState, StreamPosition failedPosition, PartitionId partition, DateTimeOffset retryTime, string reason, CancellationToken cancellationToken)
         {
-            var failingPartition = new FailingPartitionState(failedPosition, retryTime, reason, 1);
+            var failingPartition = new FailingPartitionState(failedPosition, retryTime, reason, 1, DateTimeOffset.UtcNow);
             var failingPartitions = new Dictionary<PartitionId, FailingPartitionState>(oldState.FailingPartitions)
                 {
                     [partition] = failingPartition
                 };
-            var newState = new StreamProcessorState(failedPosition + 1, failingPartitions);
+            var newState = new StreamProcessorState(failedPosition + 1, failingPartitions, DateTimeOffset.MinValue);
```
woksin, line 52: This sets the LastSuccessfullyPRocessed to MinValue. That's not right. It should be the previous value (oldState.LastSuccessfullyProcessed)

Source/Events.Processing/Streams/Partitioned/ScopedStreamProcessor.cs
```diff
@@ -54,7 +54,7 @@ protected override async Task<IStreamProcessorState> ProcessEvent(StreamEvent @e
             var streamProcessorState = currentState as StreamProcessorState;
             if (streamProcessorState.FailingPartitions.Keys.Contains(@event.Partition))
             {
-                var newState = new StreamProcessorState(@event.Position + 1, streamProcessorState.FailingPartitions);
+                var newState = new StreamProcessorState(@event.Position + 1, streamProcessorState.FailingPartitions, DateTimeOffset.MinValue);
```
woksin, line 57: The same as the above comment; This resets the LastSuccessfullyProcessed value to the minimum. It should be the last value streamProcessorState.LastSuccessfullyProcessed

PR #401, opened 2020-05-13
Add custom StreamProcessorStateDiscriminatorConvetion for AbstractStreamProcessorState serializing

joelhoisko:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1175622497640624)


joelhoisko:
Tested with the new stuff from 5.0.0, seems to be working fine. Open for review.

Source/Events.Store.MongoDB/DatabaseConnection.cs
```diff
@@ -19,6 +24,7 @@ public class DatabaseConnection
         /// <param name="configuration">A <see cref="IConfigurationFor{EventStoreConfiguration}"/> with database connection parameters.</param>
         public DatabaseConnection(IConfigurationFor<EventStoreConfiguration> configuration)
         {
+            RegisterCustomDiscriminators();
```
jakhog, line 27: Is this safe to do multiple times?

Source/Events.Store.MongoDB/Processing/Streams/AbstractStreamProcessorState.cs
```diff
@@ -10,8 +10,10 @@ namespace Dolittle.Runtime.Events.Store.MongoDB.Processing.Streams
 {
     /// <summary>
     /// Represents the base state of an <see cref="AbstractScopedStreamProcessor" />.
+    /// It has a programmatically assigned <see cref="StreamProcessorStateDiscriminatorConvention"/> which takes care of
+    /// serializing <see cref="StreamProcessorState"/> and <see cref="Partitioned.PartitionedStreamProcessorState"/> to
+    /// this collection.
```
jakhog, line 15: This comment is a bit misleading - as there really is never an instance of an `AbstractScopedStreamProcessor`. Also the last three lines should probably be a remark pointing to the `StreamProcessorStateDiscriminatorConvention` which should describe what it does. So something like:
```
The <see cref="StreamProcessorStateDiscriminatorConvention"/> is used to deserialise either a <see cref="StreamProcessorState"/> or a <see cref="Partitioned.PartitionedStreamProcessorState"/> from a stream processor state collection.
```

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,71 @@
...
+                        break;
+                    }
+                }
+
+                bsonReader.SkipValue();
+            }
+
+            bsonReader.ReturnToBookmark(bookmark);
+            return actualType;
+        }
```
jakhog, line 69: What happens in here if the `Parititioned` field is not found on the BSON document? What is the `nominalType` passed in here?

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedStateDerivedFromAbstractStreamProcessorState.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+using Dolittle.Runtime.Events.Processing.Streams;
+
+namespace Dolittle.Runtime.Events.Store.MongoDB.Processing.Streams
+{
+    /// <summary>
+    /// Exception that gets thrown when trying to use an unsuported <see cref="IStreamProcessorState"/> and <see cref="IStreamProcessorId"/> and .
+    /// </summary>
+    public class UnsupportedStateDerivedFromAbstractStreamProcessorState : Exception
```
jakhog, line 12: Should this exception extend a more generic exception defined outside the MongoDB event store implementation?

Source/Events.Store.MongoDB/DatabaseConnection.cs
joelhoisko: It's not safe to do multiple times, that's why I added the `[SingletonPerTenant]` attribute to the class so that it only get's called once per tenant.
Not sure how it deals with multiple tenants, doesn't a single tenant own a single MongoDB database with it's own mappings and registered serializers etc which wouldn't clash then?

Source/Events.Store.MongoDB/Processing/Streams/AbstractStreamProcessorState.cs
joelhoisko: Fixed the comment, the `AbstractScopedStreamProcessor` seems to be a typo I did.

Source/Events.Store.MongoDB/DatabaseConnection.cs
jakhog: Nope - they share the same code, so this would be called once for every tenant. If this thing is supposed to only be called once - I would suggest we use a static constructor (see https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors)

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: I checked and the `nominalType` is actually one of `Abstract/Partitioned/StreamProcessorState`'s so I added some logic where we can immediately return the correct type if the nominal type is also already the correct type.

But in a case of `nominalType == AbstractStreamProcessorState` and there being no `Partitioned` field the function would return `AbstractStreamProcessorState`. I probably should also make an exception for that case too...

I also added an exception dealing when the `nominalType` none of the expected types.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -33,39 +33,56 @@ class StreamProcessorStateDiscriminatorConvention : IDiscriminatorConvention
...
             {
-                var fieldName = bsonReader.ReadName();
-                if (fieldName == "Partitioned")
+                return typeof(StreamProcessorState);
+            }
+            else if (nominalType == typeof(Partitioned.PartitionedStreamProcessorState))
+            {
+                return typeof(Partitioned.PartitionedStreamProcessorState);
+            }
+            else if (nominalType == typeof(AbstractStreamProcessorState))
```
jakhog: I don't really see how this is fixing the issue. How does Mongo decide what to pass along as the `nominalType`?

And also - if there is no `Paritioned` field on the BSON document, I would expect Mongo to throw an exception elsewhere later when it is trying to instantiate an `AbstractStreamProcessorState`.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -33,39 +33,56 @@ class StreamProcessorStateDiscriminatorConvention : IDiscriminatorConvention
...
         {
-            var bookmark = bsonReader.GetBookmark();
-            bsonReader.ReadStartDocument();
-            var actualType = nominalType;
-            while (bsonReader.ReadBsonType() != BsonType.EndOfDocument)
+            if (nominalType == typeof(StreamProcessorState))
             {
-                var fieldName = bsonReader.ReadName();
-                if (fieldName == "Partitioned")
+                return typeof(StreamProcessorState);
```
jakhog, line 45: You already have the type in `nominalType`, so you could just return that...

joelhoisko: fixd

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Added exceptions for a missing `Partitioned` field.

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedStateDerivedFromAbstractStreamProcessorState.cs
joelhoisko: Removed this particular exception, some of the mongo exceptions extend the `EventStoreConsistencyError` now too

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        /// <param name="bsonReader">A <see cref="IBsonReader"/>.</param>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <returns>The actual wanted type.</returns>
+        public Type GetActualType(IBsonReader bsonReader, Type nominalType)
+        {
+            if (nominalType == typeof(StreamProcessorState) || nominalType == typeof(Partitioned.PartitionedStreamProcessorState))
+            {
+                return nominalType;
+            }
```
jakhog, line 46: To be honest - I think these cases should also be exceptions. This means that someone has registered this discriminator for any of those types, which is not right.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// <see cref="StreamProcessorState"/> and <see cref="Partitioned.StreamProcessorStateExtensions"/>.
+    /// </summary>
+    /// <remarks>
+    /// This class was based mostly on information I gathered from these 2 sources:
+    /// https://groups.google.com/forum/#!topic/mongodb-user/iOeEXbUYbo4
+    /// https://github.com/mongodb/mongo-csharp-driver/blob/6b73e381827f83af368a949a807076dad01607fc/MongoDB.DriverUnitTests/Samples/MagicDiscriminatorTests.cs#L53 .
+    /// </remarks>
+    class StreamProcessorStateDiscriminatorConvention : IDiscriminatorConvention
+    {
+        public string ElementName => "Partitioned";
```
jakhog, line 22: Make a const string with the name somewhere - so we don't have to rely on typing the same string literal twice.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            else if (nominalType == typeof(AbstractStreamProcessorState))
+            {
+                var bookmark = bsonReader.GetBookmark();
+                bsonReader.ReadStartDocument();
+                ObjectId id = default;
+                while (bsonReader.ReadBsonType() != BsonType.EndOfDocument)
+                {
+                    var fieldName = bsonReader.ReadName();
+                    if (fieldName == "Partitioned")
```
jakhog, line 55: Use const value here

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDocumentIsMissingPartitionedField.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public class StreamProcessorStateDocumentIsMissingPartitionedField : EventStoreConsistencyError
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StreamProcessorStateDocumentIsMissingPartitionedField"/> class.
+        /// </summary>
+        /// <param name="type">The given state.</param>
+        /// <param name="id">The _id field.</param>
+        public StreamProcessorStateDocumentIsMissingPartitionedField(Type type, ObjectId id)
+            : base($"Type: {type} with id: {id} didn't contain a \"Partitioned\" field.", null)
```
jakhog, line 20: Don't think it makes sense to pass along the type here, as it will always be `AbstractStreamProcessorState`. And I think it is good practice to write out the whole message here (with the text from the class name) and also the consequence of the error.

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedTypeForStreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Exception that gets thrown when trying to use an unsupported type in  <see cref="StreamProcessorStateDiscriminatorConvention"/>.
+    /// </summary>
+    public class UnsupportedTypeForStreamProcessorStateDiscriminatorConvention : EventStoreConsistencyError
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UnsupportedTypeForStreamProcessorStateDiscriminatorConvention"/> class.
+        /// </summary>
+        /// <param name="type">Nominal type used in the discriminator convention.</param>
+        public UnsupportedTypeForStreamProcessorStateDiscriminatorConvention(Type type)
+            : base($"Type: {type} isn't supported by StreamProcessorStateDiscriminatorConvention.", null)
```
jakhog, line 18: If we agree that it should only be used on `AbstractStreamProcessorState` - that should be included in this message. If we know that this has happened because it was registered using a `BsonSomething.RegisterSomething` - we could also point that out in this message.

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedTypeForStreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.Runtime.Events.Store.MongoDB.Processing.Streams
+{
+    /// <summary>
+    /// Exception that gets thrown when trying to use an unsupported type in  <see cref="StreamProcessorStateDiscriminatorConvention"/>.
+    /// </summary>
+    public class UnsupportedTypeForStreamProcessorStateDiscriminatorConvention : EventStoreConsistencyError
```
jakhog, line 11: This isn't really an event store consistency problem. This is something we have done wrong in the code. So just leave it with `: Exception`

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorState.cs
```diff
@@ -56,5 +57,18 @@ public StreamProcessorState(Guid scopeId, Guid eventProcessorId, Guid sourceStre
         /// Gets or sets a value indicating whether the Stream Processor is failing or not.
         /// </summary>
         public bool IsFailing { get; set; }
+
+        /// <summary>
+        /// Converts the <see cref="StreamProcessorState" /> to the runtime representation of <see cref="Runtime.Events.Processing.Streams.StreamProcessorState"/>.
+        /// </summary>
+        /// <returns>The converted <see cref="Runtime.Events.Processing.Streams.Partitioned.StreamProcessorState" />.</returns>
+        public override IStreamProcessorState ToRuntimeRepresentation() =>
+            new Runtime.Events.Processing.Streams.StreamProcessorState(
```
jakhog, line 66: We could have a using statement to shorten this down a bit right?

Source/Events.Store.MongoDB/Processing/Streams/Partitioned/StreamProcessorStateExtensions.cs
```diff
@@ -19,16 +19,5 @@ public static class StreamProcessorStateExtensions
         /// <returns>The converted <see cref="Runtime.Events.Processing.Streams.Partitioned.FailingPartitionState" />.</returns>
         public static Runtime.Events.Processing.Streams.Partitioned.FailingPartitionState ToRuntimeRepresentation(this FailingPartitionState state) =>
             new Runtime.Events.Processing.Streams.Partitioned.FailingPartitionState(state.Position, state.RetryTime, state.Reason, state.ProcessingAttempts, state.LastFailed);
```
jakhog, line 17: Could some usings make this a bit shorter?

Source/Events.Store.MongoDB/Processing/Streams/Partitioned/PartitionedStreamProcessorState.cs
```diff
@@ -36,5 +38,15 @@ public PartitionedStreamProcessorState(Guid scopeId, Guid eventProcessorId, Guid
         /// </summary>
         [BsonDictionaryOptions(DictionaryRepresentation.Document)]
         public IDictionary<string, FailingPartitionState> FailingPartitions { get; set; }
+
+        /// <summary>
+        /// Converts the <see cref="PartitionedStreamProcessorState" /> to the runtime representation of <see cref="Runtime.Events.Processing.Streams.Partitioned.StreamProcessorState" />.
+        /// </summary>
+        /// <returns>The converted <see cref="Runtime.Events.Processing.Streams.Partitioned.StreamProcessorState" />.</returns>
+        public override IStreamProcessorState ToRuntimeRepresentation() =>
+            new Runtime.Events.Processing.Streams.Partitioned.StreamProcessorState(
```
jakhog, line 47: Some usings to shorten?

Source/Events.Store.MongoDB/Processing/Streams/AbstractStreamProcessorState.cs
```diff
@@ -9,9 +9,13 @@
 namespace Dolittle.Runtime.Events.Store.MongoDB.Processing.Streams
 {
     /// <summary>
-    /// Represents the base state of an <see cref="AbstractScopedStreamProcessor" />.
+    /// Represents the base class for <see cref="StreamProcessorState"/> and <see cref="Partitioned.PartitionedStreamProcessorState"/>.
```
jakhog, line 12: Could we perhaps say in stead that this is the persisted version of `IStreamProcessorState` ?

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
jakhog: Ohh - just saw your reply above - that you had seen `nominalType` be those two other ones. I find that surprising. What did you do to get those types in here?

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
jakhog: Continued below...

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Just went into the `GetActualType` method with the debugger and saw that the `nominalType` was already the correct types sometimes.

The `GetActualType` method in our case is probably called from [`BsonClassMapSerializer`](https://github.com/mongodb/mongo-csharp-driver/blob/9e567e23615c8bb5c7ac1489427c2d15b2124522/src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs#L126) with the [`BsonDeserializationArgs`](https://github.com/mongodb/mongo-csharp-driver/blob/010e7ee46b085cdd3762894ece9e2d258b66ab0d/src/MongoDB.Bson/Serialization/BsonDeserializationArgs.cs) struct that just contains the `nominalType`. I couldn't find where the `BsonDeserializationArgs` get created as they're kinda used everywhere in the repo.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Put the property in there (as that's what we're actually searching for).

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDocumentIsMissingPartitionedField.cs
joelhoisko: Fixd.

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedTypeForStreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Fixd.

Source/Events.Store.MongoDB/Processing/Streams/Partitioned/StreamProcessorStateExtensions.cs
joelhoisko: Shortened.

Source/Events.Store.MongoDB/Processing/Streams/Partitioned/PartitionedStreamProcessorState.cs
joelhoisko: Shortened.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorState.cs
joelhoisko: Shortened.

Source/Events.Store.MongoDB/Processing/Streams/UnsupportedTypeForStreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Added more text to the exception.

Source/Events.Store.MongoDB/DatabaseConnection.cs
joelhoisko: Added static constructor.

Source/Events.Store.MongoDB/Processing/Streams/AbstractStreamProcessorState.cs
joelhoisko: How about
```suggestion
    /// Represents the base class for <see cref="StreamProcessorState"/> and <see cref="Partitioned.PartitionedStreamProcessorState"/>,
    /// which represent the persisted versions of <see cref="IStreamProcessorState"/>.
```

I kinda want to have a comment about the derived classes in there too.

joelhoisko: Actually I like just the "persisted version of IStreamProcessorState" more.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Removing the `[BsonKnownTypes]` fixes this and makes the `nominalType` always just `AbstractStreamProcessorState`

PR #402, opened 2020-05-13
MongoDB event store clean

woksin:
Cleans up the MongoDB event store and starts fresh with the specs. More specs will need to come and they should not be worrying about testing the actual database interaction. 

Changes:
* StreamProcessorStates => StreamProcessorStateRepository
* Moved classes related to Subscriptions to EventHorizon folder in Processing/Streams
* Removed Constants class as it proved obsolete
* Replaced some of the extension classes with the IEventConverter
* Removed all the specs that tested interactions with the database as that kind of tests should not really be present here in the Runtime-project. It should sit in a separate project that tests the database decoupled from its implementation.
* Fix broken specs in Events.Processing and Events.Store

New:
* Moved all collections and methods on collections out from EvennStoreConnection into their own classes and interfaces (IStreams, IStreamProcessorStates, ...). EventStoreConnection did way too much. I think this new split makes it much more easier to work with.
* IEventConverter : converts events back and forth between store and runtime representations
* IEventsToStreamsWriter : A system that knows how to append a new event to a stream collection. This is used in all the different event writer implementations instead of having a static method on the EventsToStreamsWriter class.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1175699932274837)


Source/Events.Store.MongoDB/Streams/IEventsToStreamsWriter.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using Dolittle.Runtime.Events.Store.Streams;
+using MongoDB.Driver;
+
+namespace Dolittle.Runtime.Events.Store.MongoDB.Streams
+{
+    /// <summary>
+    /// Defines a that can write events to a Stream.
+    /// </summary>
+    public interface IEventsToStreamsWriter
```
joelhoisko, line 15: The other interfaces are usually called `IWriteXToY` so we should probably follow the same pattern.
```suggestion
    public interface IWriteStoreEventsToStreams
```

Source/Events.Store.MongoDB/Streams/EventsToStreamsWriter.cs
```diff
@@ -14,37 +14,39 @@ namespace Dolittle.Runtime.Events.Store.MongoDB.Streams
     /// <summary>
     /// Represents an implementation of <see cref="IWriteEventsToStreams" />.
     /// </summary>
-    public class EventsToStreamsWriter : IWriteEventsToStreams
+    public class EventsToStreamsWriter : IEventsToStreamsWriter, IWriteEventsToStreams
```
joelhoisko, line 17: Renaming of the interfaces would probably makes this a bit clearer whats going on in here following the convention of adding `Store/Runtime` to the interface class/interface name depending on which they work on.

The `IWriteEventsToStreams` interface is not part of the PR so can't comment on it but it should also then get renamed.
```suggestion
    public class EventsToStreamsWriter : IWriteStoreEventsToStreams, IWriteRuntimeEventsToStreams
```

Source/Events.Store.MongoDB/EventHorizon/EventHorizonEventsWriter.cs
```diff
@@ -18,47 +17,40 @@ namespace Dolittle.Runtime.Events.Store.MongoDB.EventHorizon
...
-        /// <param name="connection">An <see cref="EventStoreConnection"/> to a MongoDB EventStore.</param>
-        /// <param name="logger">An <see cref="ILogger"/>.</param>
-        public EventHorizonEventsWriter(
-            EventStoreConnection connection,
-            ILogger logger)
+        /// <param name="streams">The <see cref="IStreams" />.</param>
+        /// <param name="eventsToStreamsWriter">The <see cref="IEventsToStreamsWriter" />.</param>
+        /// <param name="eventConverter">The <see cref="IEventConverter" />.</param>
+        /// <param name="logger">The <see cref="ILogger" />.</param>
+        public EventHorizonEventsWriter(IStreams streams, IEventsToStreamsWriter eventsToStreamsWriter, IEventConverter eventConverter, ILogger logger)
```
joelhoisko, line 32: Same thing here.

Also `IWriteEventHorizonEvents` doesn't mention anything about `Streams`, should it or is the `Streams` part always implicit in which case we could drop it from the other `IWriteXToY` interfaces?
```suggestion
        /// <param name="eventsToStreamsWriter">The <see cref="IWriteStoreEventsToStreams" />.</param>
        /// <param name="eventConverter">The <see cref="IEventConverter" />.</param>
        /// <param name="logger">The <see cref="ILogger" />.</param>
        public EventHorizonEventsWriter(IStreams streams, IWriteStoreEventsToStreams eventsToStreamsWriter, IEventConverter eventConverter, ILogger logger)
```

Source/Events.Store.MongoDB/Streams/EventsToStreamsWriter.cs
woksin: I don't like the "StoreEvents" part. I would counter suggest this:
```suggestion
    public class EventsToStreamsWriter : IWriteEventsToStreamCollection, IWriteEventsToStreams
```

Source/Events.Store.MongoDB/Streams/IEventsToStreamsWriter.cs
woksin:
```suggestion
    public interface IWriteEventsToStreamCollection
```

Source/Events.Store.MongoDB/EventHorizon/EventHorizonEventsWriter.cs
woksin:
```suggestion
        /// <param name="eventsToStreamsWriter">The <see cref="IWriteEventsToStreamCollection" />.</param>
        /// <param name="eventConverter">The <see cref="IEventConverter" />.</param>
        /// <param name="logger">The <see cref="ILogger" />.</param>
        public EventHorizonEventsWriter(IStreams streams, IWriteEventsToStreamCollection eventsToStreamsWriter, IEventConverter eventConverter, ILogger logger)
```

Source/Events.Store.MongoDB/Streams/EventsToStreamsWriter.cs
joelhoisko: Sounds good :+1:

woksin:
Your comments should be fixed by https://github.com/dolittle-runtime/Runtime/pull/402/commits/0705cb5ffba31e59605e3c236da7b51f3d262035

PR #405, opened 2020-05-18
Change DateTimeOffsets to DateTimes in EventStore

joelhoisko:
We only want to deal with DateTimeOffsets in the runtime representations but for MongoDB we want to just use DateTime's as they look prettier and are nicer to query for.

* Changed all `DateTimeOffset`s in the eventstore to use `DateTime` instead.
* `DateTimeOffset.UtcDateTime` for offset->datetime conversion
* Conversions from datetime->datetimeoffset is [implicit](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.op_implicit?view=netcore-3.1)

Do note that we're [missing out on some precision](https://alexmg.com/posts/datetime-precision-with-mongodb-and-the-c-driver) as the [ticks get squashed a tiny bit](https://stackoverflow.com/a/6039864/5806412) going from .NET to MongoDB.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1176236574870464)


joelhoisko:
Merged fresh stuff from 5.0.0 and tested, ready for review.

joelhoisko:
I kept everything in `DateTimeOffset` until the very end now and I also added `        [BsonDateTimeOptions(Kind = DateTimeKind.Utc)]` to the properties. Checked with the debugger and the read DateTime's have `Kind = UTC` on them.

Instead of marking the properties with the attributes we could also [globally specify it](https://jira.mongodb.org/browse/CSHARP-511?focusedCommentId=136390&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-136390) but I'm not sure if that's the solution here.

woksin:
@joelhoisko 

> Instead of marking the properties with the attributes we could also [globally specify it](https://jira.mongodb.org/browse/CSHARP-511?focusedCommentId=136390&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-136390) but I'm not sure if that's the solution here.

I think that we can do this. @jakhog ?

PR #406, opened 2020-05-18
Remove Scope property where scope is implicit

joelhoisko:
AbstractStreamProcessorState and SubscriptionState have a scope implicitly by being in scoped collections so we can remove it.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1176246396854066)


Source/Events.Store.MongoDB/Processing/Streams/EventHorizon/SubscriptionStates.cs
```diff
@@ -49,7 +49,6 @@ public async Task<IMongoCollection<MongoDB.Processing.Streams.EventHorizon.Subsc
                         .Ascending(_ => _.ConsumerTenantId)
```
jakhog, line 49: Why do we keep the consumer TenantId? Shouldn't that also be implicit from what event store database this is stored in?

joelhoisko: Good point, also removed `ConsumerTenantId` from SubscriptionState.

woksin:
On 3), @jakhog. We don't need the constructors. I would however suggest that we made all the classes immutable (not public setter) and instead use this [immutable type convention](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Conventions/ImmutableTypeClassMapConvention.cs).

jakhog:
But can we do that here? Isn't the whole point of these classes that they represent the document in Mongo, and that we modify the properties to change the persisted data?
Haven't looked at the code that closely, so might have gotten this wrong.

woksin:
I don't know how it's common to do this, I think that the most common way is to have public setters and no constructors and instead use the object initializer to create the objects. But I don't really like to use the object initializer for this purpose, especially when we are a team, because I'd like to argue that it's way more error prone to change the classes later because the object initializer does not care whether you initialize all the properties. For that reason I'd prefer to use a constructor, and private setter (immutable type) if possible.

joelhoisko:
I got rid of `Producer/Id` pre/post-fixes.

woksin:
> Regarding the immutability we could register our stuff with the [`ImmutableTypeClassMapConvention`](https://stackoverflow.com/a/50269397/5806412) as long as all of our properties are [read-only and present in the constructor](https://stackoverflow.com/questions/16501145/serializing-immutable-value-types-with-mongo-c-sharp-driver#comment107653421_50269397).
> 
> But maybe that task could be dealt with when we dig deeper into Bson class mappings in general later.

Yes. It's only a code quality concern, thus not really important now IMO.

PR #408, opened 2020-05-19
Resilience

woksin:
Adds some resilience to processing of events. I also couldn't help myself from fixing some log messages I came across...

Also adds a `IResilientStreamProcessorStateRepository` which is a wrapper around the stream processor state repository that wraps the calls with a policy that handles EventStoreUnavailable exception.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1176606322871606)


Source/EventHorizon/Consumer/EventProcessor.cs
```diff
@@ -49,20 +54,18 @@ public class EventProcessor : IEventProcessor
...
-            _logger.Trace($"Processing event '{@event.Type.Id}' in scope '{Scope}' from microservice '{_subscriptionId.ProducerMicroserviceId}' and tenant '{_subscriptionId.ProducerTenantId}'");
-
-            await _receivedEventsWriter.Write(@event, _consentId, Scope, cancellationToken).ConfigureAwait(false);
-            return new SuccessfulProcessing();
-        }
+        public Task<IProcessingResult> Process(CommittedEvent @event, PartitionId partitionId, CancellationToken cancellationToken) => Process(@event, cancellationToken);
 
         /// <inheritdoc/>
-        public async Task<IProcessingResult> Process(CommittedEvent @event, PartitionId partitionId, string failureReason, uint retryCount, CancellationToken cancellationToken)
+        public Task<IProcessingResult> Process(CommittedEvent @event, PartitionId partitionId, string failureReason, uint retryCount, CancellationToken cancellationToken) => Process(@event, cancellationToken);
```
jakhog, line 60: Could we maybe have some logs here writing a trace that it is retrying before calling `Process`?

Source/EventHorizon/Consumer/EventProcessor.cs
```diff
@@ -49,20 +54,18 @@ public class EventProcessor : IEventProcessor
...
+
+        async Task<IProcessingResult> Process(CommittedEvent @event, CancellationToken cancellationToken)
         {
             _logger.Trace($"Processing event '{@event.Type.Id}' in scope '{Scope}' from microservice '{_subscriptionId.ProducerMicroserviceId}' and tenant '{_subscriptionId.ProducerTenantId}'");
```
jakhog, line 64: I think this should say that something like "writing received event horizon event to scoped event log" if the writer doesn't log that, and if the writer does log that, maybe "received event from event horizon"

Source/EventHorizon/Consumer/EventProcessorPolicy.cs
```diff
@@ -0,0 +1,42 @@
...
+            _logger = logger;
+        }
+
+        /// <inheritdoc/>
+        public Type Type => typeof(EventProcessor);
+
+        /// <inheritdoc/>
+        public Polly.IAsyncPolicy Define()
+            => Polly.Policy
+                .Handle<EventStoreUnavailable>(
```
jakhog, line 34: Don't we want to catch all kinds of exceptions here?
If this policy returns without retrying, what happens to the event horizon subscription?

Source/EventHorizon/Consumer/Subscription.cs
```diff
@@ -111,6 +117,13 @@ public async Task Start()
             {
                 await _streamProcessor.Start(_cancellationToken).ConfigureAwait(false);
             }
+            catch (Exception ex)
```
jakhog, line 120: Does the subscription restart if it fails here? Since the Head won't see this error, it seems like the event horizon will just die and no one will notice.

If that is the case, we should have some retrying around this thing.

Source/EventHorizon/Consumer/Subscriptions.cs
```diff
@@ -17,18 +19,21 @@ namespace Dolittle.Runtime.EventHorizon.Consumer
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Subscriptions"/> class.
         /// </summary>
-        /// <param name="streamProcessorStates">The <see cref="IStreamProcessorStateRepository" />.</param>
+        /// <param name="streamProcessorStates">The <see cref="IResilientStreamProcessorStateRepository" />.</param>
+        /// <param name="policies">The <see cref="IPolicies" />.</param>
         /// <param name="loggerManager">The <see cref="ILoggerManager" />.</param>
-        public Subscriptions(IStreamProcessorStateRepository streamProcessorStates, ILoggerManager loggerManager)
+        public Subscriptions(IResilientStreamProcessorStateRepository streamProcessorStates, IPolicies policies, ILoggerManager loggerManager)
```
jakhog, line 33: Why not `IAsyncPolicyFor<ICanFetchEventsFromStream>` ?

Source/Events.Processing/Streams/CreateScopedStreamProcessors.cs
```diff
@@ -16,21 +17,29 @@ namespace Dolittle.Runtime.Events.Processing.Streams
...
+        /// <param name="streamProcessorStates">The <see cref="IResilientStreamProcessorStateRepository" />.</param>
         /// <param name="executionContextManager">The <see cref="IExecutionContextManager" />.</param>
+        /// <param name="policies">The <see cref="IPolicies" />.</param>
         /// <param name="loggerManager">The <see cref="ILoggerManager" />.</param>
-        public CreateScopedStreamProcessors(IEventFetchers eventFetchers, IStreamProcessorStateRepository streamProcessorStates, IExecutionContextManager executionContextManager, ILoggerManager loggerManager)
+        public CreateScopedStreamProcessors(
+            IEventFetchers eventFetchers,
+            IResilientStreamProcessorStateRepository streamProcessorStates,
+            IExecutionContextManager executionContextManager,
+            IPolicies policies,
```
jakhog, line 37: Why not `IAsyncPolicyFor<ICanFetchEventsFromStream>`?

Source/Events.Processing/Streams/Partitioned/FailingPartitions.cs
```diff
@@ -172,7 +179,8 @@ async Task<(StreamProcessorState, FailingPartitionState)> SetFailingPartitionSta
         Task<IProcessingResult> RetryProcessingEvent(FailingPartitionState failingPartitionState, CommittedEvent @event, PartitionId partition, CancellationToken cancellationToken) =>
             _eventProcessor.Process(@event, partition, failingPartitionState.Reason, failingPartitionState.ProcessingAttempts - 1, cancellationToken);
 
-        Task PersistNewState(IStreamProcessorId streamProcessorId, StreamProcessorState newState) => _streamProcessorStates.Persist(streamProcessorId, newState, CancellationToken.None);
+        Task PersistNewState(IStreamProcessorId streamProcessorId, StreamProcessorState newState) =>
+            _streamProcessorStates.Persist(streamProcessorId, newState, CancellationToken.None);
```
jakhog, line 183: This should take in a `CancellationToken` from the outside to use. Hopefully it is not used, but if we can't cancel this call, we could end up in a situation where a StreamProcessor is never stopped and therefore cannot be restarted.

Source/EventHorizon/Consumer/Subscription.cs
woksin: The ConsumerClient makes sure that the subscription is "resubscribed" if this finished or throws and exception

Source/EventHorizon/Consumer/EventProcessor.cs
woksin: https://github.com/dolittle-runtime/Runtime/pull/408/commits/4b047c18a0f2698b6f79b77036442c4daee75134

Source/EventHorizon/Consumer/EventProcessor.cs
woksin: https://github.com/dolittle-runtime/Runtime/pull/408/commits/4b047c18a0f2698b6f79b77036442c4daee75134

Source/EventHorizon/Consumer/Subscriptions.cs
woksin: https://github.com/dolittle-runtime/Runtime/pull/408/commits/554add08e459ccc031019dcbde7d23b4f6928a23

Source/Events.Processing/Streams/CreateScopedStreamProcessors.cs
woksin: https://github.com/dolittle-runtime/Runtime/pull/408/commits/554add08e459ccc031019dcbde7d23b4f6928a23

Source/Events.Processing/Streams/Partitioned/FailingPartitions.cs
woksin: https://github.com/dolittle-runtime/Runtime/pull/408/commits/619bfca5d0315b408368e016311452e94e7f6b1b

Source/EventHorizon/Consumer/Subscription.cs
jakhog: You are right - then we don't need anything here.

Source/EventHorizon/Consumer/EventProcessorPolicy.cs
jakhog: Makes sense - as long as the subscription is restarted elsewhere (as it is) then we don't need to retry other exceptions here.

The last one was probably a glitch in my brain, I was thinking that exceptions that weren't handled here would dissapear. But I assume then that anything other than `EventStoreUnavailable` will bubble out from the policy, and then be handled elsewhere.

PR #411, opened 2020-05-20
Create missing policy

woksin:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1176788883887604)

### DIFF
--- a/Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
+++ b/Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
@@ -38,6 +38,6 @@ namespace Dolittle.Runtime.Events.Processing.Streams
                         _logger.Error(_, "Could not persist stream processor state to the event store, will retry in one second.");
                         return true;
                     })
-                .WaitAndRetryForeverAsync(attempt => TimeSpan.FromSeconds(Math.Min(Math.Pow(2, attempt), 10)));
+                .WaitAndRetryForeverAsync(attempt => TimeSpan.FromSeconds(1);
     }
 }
--- a/Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
+++ b/Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
@@ -4,8 +4,6 @@
 using System;
 using Dolittle.Logging;
 using Dolittle.Resilience;
-using Dolittle.Runtime.Events.Store;
-using Dolittle.Runtime.Events.Store.Streams;
 using Polly;
 
 namespace Dolittle.Runtime.Events.Processing.Streams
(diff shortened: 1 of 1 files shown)


Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public class ResilientStreamProcessorStateRepositoryPolicy : IDefineAsyncPolicyForType
+    {
+        readonly ILogger _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ResilientStreamProcessorStateRepositoryPolicy"/> class.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger" />.</param>
+        public ResilientStreamProcessorStateRepositoryPolicy(ILogger<ICanFetchEventsFromStream> logger)
```
jakhog, line 24: Wrong kind of logger.
```suggestion
        public ResilientStreamProcessorStateRepositoryPolicy(ILogger<ResilientStreamProcessorStateRepository> logger)
```

Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
```diff
@@ -0,0 +1,43 @@
...
+            _logger = logger;
+        }
+
+        /// <inheritdoc/>
+        public Type Type => typeof(ResilientStreamProcessorStateRepository);
+
+        /// <inheritdoc/>
+        public Polly.IAsyncPolicy Define() =>
+            Polly.Policy
+                .Handle<EventStoreUnavailable>(
```
jakhog, line 35: This policy should handle `Exception`, it should forever no matter what happens.
```suggestion
                .Handle<Exception>(
```

Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
```diff
@@ -0,0 +1,43 @@
...
+                    _ =>
+                    {
+                        _logger.Debug(_, "Event Store is unavailable");
```
jakhog, line 38: These exceptions are quite serious, so I suggest we make use of `.Error(...)` here, and give the log a bit more context.
```suggestion
                        _logger.Error(_, "Could not persist stream processor state to the event store, will retry in one second.");
```

Source/Events.Processing/Streams/ResilientStreamProcessorStateRepositoryPolicy.cs
```diff
@@ -0,0 +1,43 @@
...
+                        return true;
+                    })
+                .WaitAndRetryForeverAsync(attempt => TimeSpan.FromSeconds(Math.Min(Math.Pow(2, attempt), 10)));
```
jakhog, line 41: I suggest we retry this thing more often, as we want to try our very best to make this happen before the Runtime is stopped for whatever reason (and this state is lost). Also, if these exceptions stem from e.g. Mongo connection errors, they already have a significant timeout on them.
```suggestion
                .WaitAndRetryForeverAsync(attempt => TimeSpan.FromSeconds(1);
```

woksin:
Thanks for the suggested changes, @jakhog. I see that I was way to quick on the trigger this time as well...

PR #413, opened 2020-05-22
AbstractFilterDefinition discriminator

joelhoisko:
- Reworks `AbstractFilterDefinition` to look more like `IFilterDefinition` (see pic attached)
- Adds a custom discriminator for `AbstractFilterDefinition` which decides the type between `FilterDefinition/PublicFilterDefinition/TypePartitionFilterDefinition`.
- Exceptions for the discriminator.

V5 doodles

Wasn't sure how `ThrowIfFieldsNotFound` should work in cases where neither of the values are set...

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1177062802315483)


woksin:
It seems good, but I'm not entirely sure on the enum thingy

joelhoisko:
The idea is that in the future we might add different filters that work in the say way as `TypePartitionFilterDefinition` as in declarative filters that run in the runtime that might filter not only depending on the `Type` with other things too like values etc. @jakhog knows more about these grand plans.

woksin:
Yes, I understand. It's just the the Enum mechanism is kinda breaking the Open-Closed Principle and I'm thinking if there's a better way to do this particular thing

jakhog:
I obviously still haven't understood this _open-closed principle_ since I don't understand what you mean in this case. But keep in mind that the purpose of the changes in this PR is about the persisted data in Mongo, and decoupling that from the csharp implementations.

So effectively what this code has done is changing the persisted filter definitons from something like
```
{
   _t: [AbstractFilterDefinition, PublicFilterDefinition]
  ...
}
```
to
```
{
   Type: "Remote"
   Public: true
   ...
}
```

So, this whole exercise was to allow us to change the underlying code - e.g. to be more (or less) open closed - without it affecting the persisted data.

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
```diff
@@ -33,7 +33,7 @@ class StreamProcessorStateDiscriminatorConvention : IDiscriminatorConvention
 
         /// <summary>
         /// Gets the correct type when deserializing objects from  <see cref="AbstractStreamProcessorState"/> collection
-        /// depending on the given nominal type or "Partitioned" field.
+        /// depending on the "Partitioned" field. Also saves the _id field for the exception if needed.
```
jakhog, line 36: What does "Also saves the _id field for the exception if needed." mean?

If you are reffering to the `ObjectId` that is stored in the loop to throw meaningful exceptions, this should not be described in the documentation summary, as it is purely internal to the method.

Source/Events.Store.MongoDB/Streams/Filters/AbstractFilterDefinition.cs
```diff
@@ -36,6 +38,16 @@ protected AbstractFilterDefinition(Guid filterId, Guid sourceStream)
         /// </summary>
         public Guid SourceStream { get; set; }
```
jakhog, line 39: I believe this property should not be present on the filter definition, as it always has to be the Event Log.

Source/Events.Store.MongoDB/Streams/Filters/AbstractFilterDefinitionDocumentIsMissingAField.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class AbstractFilterDefinitionDocumentIsMissingAField : EventStoreConsistencyError
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AbstractFilterDefinitionDocumentIsMissingAField"/> class.
+        /// </summary>
+        /// <param name="id">The _id field.</param>
+        /// <param name="field">The missing field.</param>
+        public AbstractFilterDefinitionDocumentIsMissingAField(Guid id, string field)
+            : base($"The {field} was missing from the document with id: {id}", null)
```
jakhog, line 19:
```suggestion
            : base($"The {field} was missing from the filter definition document with id: {id}", null)
```

Source/Events.Store.MongoDB/Streams/Filters/AbstractFilterDefinitionDocumentIsMissingAField.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.Runtime.Events.Store.MongoDB.Streams.Filters
+{
+    /// <summary>
+    /// Exception that gets thrown when a field is missing from <see cref="AbstractFilterDefinition"/> document.
```
jakhog, line 9:
```suggestion
    /// Exception that gets thrown when a required field is missing from <see cref="AbstractFilterDefinition"/> document.
```

Source/Events.Store.MongoDB/Streams/Filters/FilterDefinitionDiscriminatorConvention.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            /// <summary>
+            /// Represents filters that get filtered in the runtime depending on a given declaration.
+            /// </summary>
+            EventTypeId
+        }
+
+        /// <summary>
+        /// Gets the fields name in mongo.
+        /// </summary>
```
jakhog, line 39:
```suggestion
        /// <inheritdoc/>
```

Source/Events.Store.MongoDB/Streams/Filters/FilterDefinitionDiscriminatorConvention.cs
```diff
@@ -0,0 +1,133 @@
...
+        public string ElementName => Type;
+
+        /// <summary>
+        /// Gets the value to be put into the ElementName field. The enum is saved as a string so that it looks prettier
+        /// when inspecting the db.
+        /// </summary>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <param name="actualType">The actual type.</param>
+        /// <returns>A string.</returns>
```
jakhog, line 48:
```suggestion
        /// <inheritdoc/>
```

Source/Events.Store.MongoDB/Streams/Filters/FilterDefinitionDiscriminatorConvention.cs
```diff
@@ -0,0 +1,133 @@
...
+                throw new UnsupportedTypeForFilterDefinitionDiscriminatorConvention(actualType);
+            }
+        }
+
+        /// <summary>
+        /// Gets the correct type when deserializing objects from  <see cref="AbstractFilterDefinition"/> collection.
+        /// </summary>
+        /// <param name="bsonReader">A <see cref="IBsonReader"/>.</param>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <returns>The actual wanted type.</returns>
```
jakhog, line 70:
```suggestion
        /// <inheritdoc/>
```

Source/Events.Store.MongoDB/Streams/Filters/FilterDefinitionDiscriminatorConvention.cs
```diff
@@ -0,0 +1,133 @@
...
+                    case Type:
+                        filterType = Enum.Parse<FilterType>(bsonReader.ReadString());
+                        isFilterSet = true;
+                        break;
+                    case Public:
+                        publicValue = bsonReader.ReadBoolean();
+                        isPublicSet = true;
+                        break;
+                    case "_id":
+                        id = (Guid)bsonReader.ReadBinaryData();
```
jakhog, line 95: This seems a bit dangerous, as there is special things setup in the MongoClient for how to deal with `Guid`s.

Source/Events.Store.MongoDB/Streams/Filters/UnsupportedFilterTypeEnumValue.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class UnsupportedFilterTypeEnumValue : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UnsupportedFilterTypeEnumValue"/> class.
+        /// </summary>
+        /// <param name="filterType">string of the given FilterType.</param>
+        /// <param name="id">Id of the document.</param>
+        public UnsupportedFilterTypeEnumValue(FilterDefinitionDiscriminatorConvention.FilterType filterType, Guid id)
+            : base($"Document id: {id} has an unsupported FilterType: {filterType}. Check supportd types from {typeof(FilterDefinitionDiscriminatorConvention)}.")
```
jakhog, line 19:
```suggestion
            : base($"Filter definition document id: {id} has an unsupported FilterType: {filterType}. Supported types are \"Remote\", \"EventTypeId\".")
```

Source/Events.Store.MongoDB/Streams/Filters/UnsupportedFilterTypeEnumValue.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.Runtime.Events.Store.MongoDB.Streams.Filters
+{
+    /// <summary>
+    /// Exception that gets thrown when a <see cref="AbstractFilterDefinition"/> document has an unsupported Type field.
+    /// </summary>
+    public class UnsupportedFilterTypeEnumValue : Exception
```
jakhog, line 11:
```suggestion
    public class UnsupportedFilterTypeEnumValue : EventStoreConsistencyError
```

Source/Events.Store.MongoDB/Processing/Streams/StreamProcessorStateDiscriminatorConvention.cs
joelhoisko: Fixd, made it just `inheritdoc`

Source/Events.Store.MongoDB/Streams/Filters/UnsupportedFilterTypeEnumValue.cs
joelhoisko: Fixed and added a `null` as the given innerException to the constructor.

Source/Events.Store.MongoDB/Streams/Filters/FilterDefinitionDiscriminatorConvention.cs
jakhog: This method is probably safer - passed along with the same GuidRepresentation that we use in the client (which should be the standard one).
https://mongodb.github.io/mongo-csharp-driver/2.5/apidocs/html/M_MongoDB_Bson_BsonBinaryData_ToGuid_1.htm

Source/Events.Store.MongoDB/Streams/Filters/AbstractFilterDefinition.cs
joelhoisko: Removed all of the properties from `AbstractFilterDefinitions`.

PR #417, opened 2020-05-28
Docs for configuration files

joelhoisko:
WIP

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1177811056688713)

### DIFF
--- a/Documentation/setup_configuration.md
+++ b/Documentation/setup_configuration.md
@@ -1,0 +1,148 @@
+---
+title: Setup configuration
+description: Setup runtime with correct configuration files
+keywords: setup, configuration, json, resources, tenancy
+author: joel
+---
+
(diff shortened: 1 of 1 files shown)


jakhog:
I was going to suggest splitting things up a bit here, but it quickly got very complicated. So here is my suggestion:

Make a table listing all the different config files, and have columns for where they should be used, if they are required, and if they are provided by Dolittle when deployed in the platform.

So something like:

| Configuration | Head | Runtime | Required | Provided |
|---------------|------|----------|----------|----------|
| tenants.json | x | x | x | x |
| clients.json | x |   |   | x |

Then we can describe each file once. Does that make sense?

woksin:
Should we have a list for each of the configs describing each of the "<>" variables?

jakhog (approved):

PR #418, opened 2020-05-28
Fix logging to use new format

joelhoisko:
Tested in triple trouble with trace level logging on, no `FormatExceptions` in the debugger.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1177839404670112)

### DIFF
--- a/Source/Events.Processing/Filters/FilterValidators.cs
+++ b/Source/Events.Processing/Filters/FilterValidators.cs
@@ -61,7 +61,7 @@ namespace Dolittle.Runtime.Events.Processing.Filters
             {
                 if (TryGetValidatorTypeFor(filterDefinitionType, out var validatorType))
                 {
-                    _logger.Trace("Filter definition type {FilterFullName} can be validated by validator type {ValidatorFullName}", filterDefinitionType.FullName, validatorType.FullName);
+                    _logger.Trace("Filter definition type {FilterType} can be validated by validator type {ValidatorType}", filterDefinitionType, validatorType);
                     _filterDefinitionToValidatorMap.TryAdd(filterDefinitionType, validatorType);
                 }
             });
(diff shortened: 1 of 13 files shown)


Source/Events.Processing/Filters/FilterValidators.cs
```diff
@@ -61,7 +61,7 @@ void PopulateFilterValidatorMap()
             {
                 if (TryGetValidatorTypeFor(filterDefinitionType, out var validatorType))
                 {
-                    _logger.Trace($"Filter definition type {filterDefinitionType.FullName} can be validated by validator type {validatorType.FullName}");
+                    _logger.Trace("Filter definition type {FilterFullName} can be validated by validator type {ValidatorFullName}", filterDefinitionType.FullName, validatorType.FullName);
```
woksin, line 64: We should drop the .FullName here
```suggestion
                    _logger.Trace("Filter definition type {FilterType} can be validated by validator type {ValidatorType}", filterDefinitionType, validatorType);
```

Source/Events.Processing/Filters/FilterValidators.cs
```diff
@@ -74,7 +74,10 @@ bool TryGetValidatorTypeFor(Type filterDefinitionType, out Type validatorType)
             {
                 if (implementations.Count() > 1)
                 {
-                    _logger.Warning($"There are multiple validators that can validate filter defintion of type {filterDefinitionType.FullName}:\n{string.Join("\n", implementations.Select(_ => _.FullName))}\nUsing the first validator.");
+                    _logger.Warning(
+                        "There are multiple validators that can validate filter defintion of type {FullName}:\n{Implementations}\nUsing the first validator.",
+                        filterDefinitionType.FullName,
+                        string.Join("\n", implementations.Select(_ => _.FullName)));
```
woksin, line 80:
```suggestion
                    _logger.Warning(
                        "There are multiple validators that can validate filter definition of type {FilterDefinitionType}:\n{ImplementationTypes}\nUsing the first validator.",
                        filterDefinitionType,
                        string.Join("\n", implementations.Select(_ => _.ToString())));
```

joelhoisko:
Added the changes 👍

PR #421, opened 2020-06-24
Add ping mechanism

woksin:
Depends on https://github.com/dolittle/DotNET.Fundamentals/pull/320

* Changes ConsumerService and ConsumerClient to use the reverse call mechanism
* Change reverse calls to work with the new signatures

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1181795555770859)


Source/EventHorizon/Consumer/ConsumerClient.cs
```diff
@@ -141,7 +142,32 @@ protected virtual void Dispose(bool disposeManagedResources)
...
+                (message, arguments) => message.SubscriptionRequest = arguments,
+                message => message.SubscriptionResponse,
+                message => message.Request,
+                (message, response) => message.Response = response,
+                (arguments, context) => arguments.CallContext = context,
+                request => request.CallContext,
+                (response, context) => response.CallContext = context,
+                message => message.Ping,
+                (message, pong) => message.Pong = pong,
+                TimeSpan.FromSeconds(7));
```
joelhoisko, line 163: The SDK seems to use 5 seconds for the ping interval, should this also be 5 or is the extra 2 seconds for the server to have time to catch up to?

woksin: That was the intention. I think it's fair to have a little slower ping interval on event horizon

PR #422, opened 2020-07-02
Fix event handler cancellation

joelhoisko:
We had a bug with all of the StreamProcessors not shutting down properly, causing the `EventHandlersService.Connect()` method to never end.
We never caught exceptions in the `TryGetExceptions()` call so we never
cancelled the `linkedTokenSource` meaning the code would hang forever on
the `WhenAll()`.

This PR fixes this by cancelling the `linkedTokenSource` after any of the StreamProcessors or the dispatcher finishes running.
Also adds more logging.

Added:
* More Logging
* EventHorizon SubscriptionFailures.SubscriptionCancelled FailureId
* Catch 'TaskCancelledException' when attempting to subscribe to an event horizon. This exception can occur when the subscription is cancelled by client or server before it actually manages to connect to the other runtime for the event horizon. Returns a failed subscription response
* Use IHostApplicationLifetime in Event.Processing services to successfully cancel all running tasks when server shuts down. There was a problem where the background task that pings the client was never killed / finished / cancelled when the server was killed by a SIGINT. I don't actually like this that much, but I could not find any better alternatives

Changed:
* Overall logging improvements
* Cancel linked token source in Events.Processing services before waiting for all tasks to finish to make sure that they are actually cancelled
Fixed:
* Used Exception.InnerException in a bad way many places which could result in null-references

### DIFF
--- a/Source/Events.Processing/EventHandlers/EventHandlersService.cs
+++ b/Source/Events.Processing/EventHandlers/EventHandlersService.cs
@@ -102,13 +102,14 @@ namespace Dolittle.Runtime.Events.Processing.EventHandlers
 
             var sourceStream = StreamId.EventLog;
             var eventHandlerId = arguments.EventHandlerId.To<EventProcessorId>();
+            _logger.Debug("Connecting Event Handler '{EventHandlerId}'", eventHandlerId);
             StreamId targetStream = eventHandlerId.Value;
             var scopeId = arguments.ScopeId.To<ScopeId>();
             var types = arguments.Types_.Select(_ => _.Id.To<ArtifactId>());
             var partitioned = arguments.Partitioned;
             if (targetStream.IsNonWriteable)
             {
-                _logger.Warning("Cannot register Event Handler: '{eventHandlerId}' because it is an invalid Stream Id", eventHandlerId);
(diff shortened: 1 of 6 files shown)


Source/Events.Processing/EventHandlers/EventHandlersService.cs
```diff
@@ -196,37 +196,44 @@ public class EventHandlersService : EventHandlersBase
...
+                _logger.Warning(ex, "An error occurred while processing Event Handler: '{EventHandlerId}' in Scope: '{ScopeId}'", eventHandlerId, scopeId);
+                linkedTokenSource.Cancel();
                 await Task.WhenAll(tasks).ConfigureAwait(false);
                 ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
             }
+            else
+            {
+                linkedTokenSource.Cancel();
+                await Task.WhenAll(tasks).ConfigureAwait(false);
+            }
```
woksin, line 227: Nice catch. To my eyes the missing piece here was the linkedTokenSource.Cancel() call beneath the if-block. And since we throw an exception at the end of the if-block we don't actually need the else-block.

woksin, line 227: The other 2 services also needs this. I will do this in them as well

woksin:
Sorry for hijacking this PR... :)

PR #435, opened 2020-10-10
Fix JSON <-> BSON conversion in MongoDB event store

jakhog:
This PR fixes the issue of long numbers being encoded in a special way in MongoDB, and not being decoded correctly when using the `BsonDocument.ToString()` method.

I also split out the conversion as its own interface and class to make the testing more easy, and keep the logic in its own place (also allows future improvements). There is a new specification that covers the discovered bug, plus quite a few other possible problems.

While working on it, I realised that we don't support any kind of JSON as content when we convert the document to BSON for storage in MongoDB. E.g.:
- Duplicate keys are allowed in JSON, but not in BSON
- JSON allows arbitrary precision numbers, which BSON does not

While these cases are not an actual problem for almost all practical cases, we and users need to be aware of them. So we need to document this somehow, and perhaps consider an alternate approach when we encounter things that cannot be converted to BSON properly.

I'll leave that for future work.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1198181460331755)


woksin (approved):

woksin:
@jakhog Remember to label the PR. Is it a patch?

jakhog:
Yes, thanks 🙂

PR #446, opened 2020-12-28
Write about streams and event processors

woksin:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199374517701650)


Documentation/Concepts/event_handlers_and_filters.md
```diff
@@ -0,0 +1,95 @@
...
+* The ordering of the events cannot change
+* Events can only be appended to the end of the stream
+* Events cannot be removed from the stream
+* A partitioned stream cannot be changed to be unpartitioned and vice versa
+
+The Runtime will enforce these rules. You're not allowed to do anything that breaks these rules.
+
+## Processors
+
+If you're a bit confused by now don't worry, you're not alone. The word "event processor" is used many places with slightly different semantics. Stream processors are internal constructs and the kinds of event processors we'll be talking about now, however, are strictly the things that the applications will be interacting with.
```
joelhoisko, line 65: So as the quip here says, it is a bit confusing.

My suggestion for reworking this whole page:

- split out the streams part into it's own page, there's well enough content there for it's own stuff
- On the page where you explain Event Handlers and FIlters you can say that they are what we call "event processors" and link to the 2 different types of event processors we have and leave it at that. if the reader wants to really learn about stream processors then they can read about them.
- Mention that event handlers is a combination of 2 processors (filter and then the processors) and that they create streams (once again linking to the streams page)
- Images would do wonders, maybe we can make some cool ones tomorrow

joelhoisko (changes requested):
Nice, lotsa text. I added a few suggestions for simplifying the text and also a more grand suggestion of splitting this to 2 pages, streams and event handlers & filters.

joelhoisko (approved):

PR #448, opened 2020-12-29
Documentation events

joelhoisko:
I'd like some extra comments especially about the code example in the _Generations_ sections. I don't think we really have a nice way of demonstrating how to handle different generations of event types (minus the problems of having 2 types with the same name). Maybe the section regarding the generation should just be scrapped as you can't really use them right now?

Also fixed some typos on streams etc.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199379624809611)


Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+For the whole structure of an event as defined in the gRPC, please check [Contracts](https://github.com/dolittle/Contracts/tree/master/Source/Runtime/Events).
+
+### Content
+This is the object the user commits.
+
+### EventSourceId
+`EventSourceId` represents the source of the event like a "primary key" in a traditional database. By default, [partitioned event handlers]({{< ref "event_handlers_and_filters#event-handlers" >}}) use it as the `PartitionId`.
+
+### EventType
+An `EventType` is a combination of an `EventTypeId` to uniquely identify the type of event it is and the event types `Generation`.
```
woksin, line 67: Little grammar fix and suggestion
```suggestion
An `EventType` is the combination of an `EventTypeId` to uniquely identify the type of event it is and the event type's `Generation`.
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+This decouples the event from a programming language and enables the renaming of events as the domain language evolves.
+Simply put, event types are a wrapper for the actual type of your event. They mask the underlying language-specific type implementation and conventions.
```
woksin, line 69: The first sentence was, to me, a bit confusing. I think that we could just drop these two sentences as it seems to me as that it basically says the same thing as the sentence above
```suggestion
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+The Runtime doesn't know or care about the event's content, properties, or type (in its respective programming language). The Runtime saves the event to the event store and then calls the respective [EventHandlers & Filters]({{< ref "event_handlers_and_filters" >}}). For this event to be serialized to JSON and then deserialized back to a type that the client's filters and event handlers understand, an event type is required.
```
woksin, line 71: I would formalise the "event store part" a bit differently.
```suggestion
The Runtime doesn't know or care about the event's content, properties, or type (in its respective programming language). The Runtime saves the event to the event log and from that point the event is ready to be processed. You can read more about how events are processed in the section about [EventHandlers & Filters]({{< ref "event_handlers_and_filters" >}}). For this event to be serialized to JSON and then deserialized back to a type that the client's filters and event handlers understand, an event type is required.
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+{{< alert title="Why not use class/type names instead of GUIDs?" color="info" >}}
+When consuming events from other microservices it's important to remember that they name things according to their own domain and conventions.
+
+As an extreme example, a microservice could have an event with a type `CustomerRegistered`. But in another microservice in a different domain, written in a different language, this event type could be called `user_added`.
+
+GUIDs also solve the problem of having duplicate names, it's not hard to imagine having to have multiple events with the type of `CustomerRegisterd` in your code coming from different microservices.
+{{< /alert >}}
+
+#### Generations
+As the code changes, the event type is also bound to change at some point. These iterations on the same event type are called _generations_. Whenever you add or change a property in an event you should increase this number. This way the filters and handlers can handle a specific generation of an event.
```
woksin, line 88: I would say the structure of the event, rather than the evnet type
```suggestion
As the code changes, the structures and contents of your events are also bound to change at some point. In most scenarios you will see that you have the need to add more information to events. These iterations on the same event type are called _generations_. Whenever you add or change a property in an event you should increase this number. This way the filters and handlers can handle a specific generation of an event.
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+For example, imagine an older generation of an event type didn't have a property. You can have a separate handler for that particular generation that handles the missing property.
```
woksin, line 90: I don't know if we should have this here yet since we're not quite sure how we want this look look like and how to solve it yet. I suggest just removing it

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+[Handle("1844473f-d714-4327-8b7f-5b3c2bdfc26a", 2)]
+Handle(DishPrepared @event) { ... }
+```
+
+{{< alert title="Production ready" color="info" >}}
+For making development easier, you shouldn't worry about incrementing the generation until you're in production.
+{{< /alert >}}
+
+### Public vs. Private
+There is a basic distinction between private events and public events. In much the same way that you would not grant access to other applications to your internal database, you do not allow other applications to subscribe to your private events.
```
woksin, line 120: Minor correction suggestion. You subscribe to streams, not events.
```suggestion
There is a basic distinction between private events and public events. In much the same way that you would not grant access to other applications to your internal database, you do not allow other applications to receive any of your private events.
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+Private events are only accessible within a single [Tenant]({{< ref "tenant" >}}) so that an event committed for one tenant cannot be handled outside of that tenant. Private events model the system within its domain and should be named in a way that makes sense inside the domain.
```
woksin, line 122: I guess that the last sentence here is true for both private and public events? That's at least my approach to this. I would suggest just removing the last sentence.
```suggestion
Private events are only accessible within a single [Tenant]({{< ref "tenant" >}}) so that an event committed for one tenant cannot be handled outside of that tenant.
```

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+Public events are also accessible within a single tenant but they can also be added to a public [Stream]({{< ref "event-handlers" >}}) for other microservices to consume.
+
+Your external stream of public events is your contract to the outside world, your API.
```
woksin, line 126: I would just say public stream at this point and link to the section that I will write shortly :) 
```suggestion
Your [public event streams]({{< ref "streams.md#public-streams" >}}) is your contract to the outside world, your API.
```

woksin, line 126: To me the "contract to the outside world"-part is maybe a bit inaccurate? Contract seems a bit off

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+{{< alert title="Changes to public events" color="primary" >}}
+Extra caution should be paid to changing public events so as not to break other microservices consuming those events.
```
woksin, line 129: So this is the current situation, but in the future we'd have strategies for migrations that will hopefully solve this. As this statement is rather concerning, if you changing code in your domain can potentially break another Microservice. That is scary and not exactly something we want to happen

woksin: Should we say something about that here? What do you think

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+{{< /alert >}}
+
+## Commit vs Publish
+We use the word `Commit` rather than `Publish` when talking about saving events to the [Event Store]({{< ref "event_store" >}}). We want to emphasize that it's the event store that is the source of truth in the system. The act of calling filters/event handlers comes _after_ the event has been committed to the event store.
```
woksin, line 133: To add my 0.02$ here I would also talk about how when talking about "Publish" you'd usually think about publishing events to specific streams (or Topics) and event handlers, microservices etc. While that is not the case for us

woksin (changes requested):
Good stuff. I have a couple of comments and suggestions though.

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Events
+description: The source of truth in the system
```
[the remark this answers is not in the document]
joelhoisko: Yeah, system is vague. In the old text we basically called everything a "domain", I mostly replaced all of those with "system". Maybe system could be quickly defined somewhere at the top? Or maybe it's clear enough

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+- ✅ `DishAddedToMenu`
+- ❌ `TakeoutServerReady`
+- ❌ `MenuListingElementUpdated`
+
+## Main structure of an Event
+This is a simplified structure of the main parts of an event.
+
+```csharp
+Event {
+    Content object
```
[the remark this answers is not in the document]
joelhoisko: Added more clarifications for the JSON-stringines.

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+        Generation int
+    }
+    Public bool
+}
+```
+
+For the whole structure of an event as defined in the gRPC, please check [Contracts](https://github.com/dolittle/Contracts/tree/master/Source/Runtime/Events).
+
+### Content
+This is the object the user commits.
```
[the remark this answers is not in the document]
joelhoisko: Added a line bout JSON serialization

Documentation/Concepts/events.md
```diff
@@ -0,0 +1,133 @@
...
+
+### EventSourceId
+`EventSourceId` represents the source of the event like a "primary key" in a traditional database. By default, [partitioned event handlers]({{< ref "event_handlers_and_filters#event-handlers" >}}) use it as the `PartitionId`.
```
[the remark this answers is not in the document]
joelhoisko:
```suggestion
`EventSourceId` represents the source of the event like a "primary key" in a traditional database. By default, [partitioned event handlers]({{< ref "event_handlers_and_filters#event-handlers" >}}) use it for [partitioning]({{< ref "streams.md#partitions" >}}).
```
Is just a link enough?

Documentation/Concepts/events.md
joelhoisko: Fixed it and added just a link for it.

Documentation/Concepts/events.md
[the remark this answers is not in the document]
woksin: I don't know. It just felt weird to mention a specific thing like PartitionId here

Documentation/Concepts/events.md
joelhoisko: Added a warning and removed the code example

Documentation/Concepts/events.md
joelhoisko: Fixed typo in your suggestion and added it :+1:

joelhoisko: Changed the line.

Documentation/Concepts/events.md
joelhoisko: Added a line saying 'we'r working on eeet'

Documentation/Concepts/events.md
joelhoisko: Added some extra sentences explaining this and the "globalness" of committing an event

Documentation/Concepts/events.md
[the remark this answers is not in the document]
woksin: We can keep it as is for now

woksin (approved):

PR #451, opened 2021-01-04
Tenants docs

joelhoisko:
A simple writeup about tenants and multi-tenancy. Not quite sure what to say here tbh, maybe there isn't just a lot to say except  that multi-tenancy is a thing and we support it out of the box? I feel like a tutorial showing multi tenancy stuff would be more useful. Or maybe just having code samples here, but then we run back into the question of who owns the docs...

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199388975490015)

### DIFF
--- a/Documentation/Concepts/tenants.md
+++ b/Documentation/Concepts/tenants.md
@@ -4,7 +4,7 @@ description: What is a Tenant & Multi-tenancy
 weight: 13
 ---
 
-Dolittle supports defining multiple tenants using the same software out of the box.
+Dolittle supports having multiple tenants using the same software out of the box.
 
 ## What is a Tenant?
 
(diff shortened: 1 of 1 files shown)


Documentation/Concepts/tenants.md
```diff
@@ -0,0 +1,23 @@
...
+title: Tenants
+description: What is a Tenant & Multi-tenancy
+weight: 13
+---
+
+Dolittle supports defining multiple tenants using the same software out of the box.
+
+## What is a Tenant?
+
+A Tenant is a single client that's using the hosted software and infrastructure. In a SaaS (Software-as-a-Service) domain, a tenant would be a single customer using the service. The tenant has its privileges and resources only it has access to.
```
woksin, line 11:
```suggestion
A Tenant is a single client that's using the hosted software and infrastructure. In a SaaS (Software-as-a-Service) domain, a tenant would usually be a single customer using the service. The tenant has its privileges and resources only it has access to.
```

woksin (approved):

PR #455, opened 2021-01-11
Event Store documentation

joelhoisko:
- Added a big ol list of the whole structure of the MongoDB implementation of the event store. Would be nice if this could be like a generated thing or we could link to our runtime c# api docs to show the classes tha build them. Alas, this is quite nice for the users for now.
- Moved commit-vs-publish line to the Event Store page from Events.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199539487825833)

### DIFF
--- a/Documentation/Concepts/event_store.md
+++ b/Documentation/Concepts/event_store.md
@@ -10,11 +10,13 @@ aliases:
 
 An Event Store is a database optimized for storing events in an [Event Sourced]({{< ref "event_sourcing" >}}) system. The [Runtime]({{< ref "overview" >}}) manages the connections and structure of the stored data.
 
+Each [Tenant]({{< ref "tenants" >}}) has their own event store database, which is configured in [`resources.json`]({{< ref "docs/reference/runtime/configuration#resourcesjson" >}}).
+
 ## Structure of the Event Store
 
 {{< tabs name="collections" >}}
 {{% tab name="MongoDB" %}}
-This is the structure of the event store implemented in [MongoDB](https://www.mongodb.com/) in a database called `event_store`. It includes several collections, which are built as needed:
+This is the structure of the event store implemented in [MongoDB](https://www.mongodb.com/). It includes the following collections in the default [Scope]({{< ref "#scope" >}}):
 
 - `event-log`
 - `aggregates`
@@ -23,17 +25,11 @@ This is the structure of the event store implemented in [MongoDB](https://www.mo
 - `stream-<streamID>`
 - `public-stream-<streamID>`
 
-Collections can also be [scoped]({{< ref "#scope" >}}), which is reflected in their naming:
 
-- `x-<scopeID>-event-log`
-- `x-<scopeID>-stream-definitions`
-- `x-<scopeID>-stream-processor-states`
-- `x-<scopeID>-stream-<streamID>`
-- `x-<scopeID>-subscription-states`
-
-Following JSON structure examples have each fields [BSON type](https://docs.mongodb.com/manual/reference/bson-types/) as the value.
+Following JSON structure examples have each property's [BSON type](https://docs.mongodb.com/manual/reference/bson-types/) as the value.
 
 ### `event-log`
+
 The Event Log includes all the [Events]({{< ref "events" >}}) committed to the event store in chronological order. [Aggregate]({{< ref "aggregates" >}}) events have `"wasAppliedByAggregate":  true` set and events coming over the [Event Horizon]({{< ref "event_horizon" >}}) have `"FromEventHorizon": true"` set.
 
 This is the structure of a committed event:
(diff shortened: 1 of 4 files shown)


Documentation/Concepts/event_store.md
```diff
@@ -2,90 +2,189 @@
...
-
-[to be added]
-
-
-## Event Migration
-
-[to be added]
+An Event Store is a database optimized for storing [Events]({{< ref "events" >}}) in an [Event Sourced]({{< ref "event_sourcing" >}}) system. The [Runtime]({{< ref "overview" >}}) manages the connections and structure of the stored data. All [Streams]({{< ref "streams" >}}), [Event Handlers & Filters]({{< ref "event_handlers_and_filters" >}}), [Aggregates]({{< ref "aggregates" >}}) and [Event Horizon Subscriptions]({{< ref "event_horizon#subscription" >}}) are being kept track inside the event store.
+
+The event store is an append-only database. Any events saved to it **cannot be changed** or **deleted**. It acts as the record of all events that have happened in the system from the beginning of time.
```
woksin, line 11: This may or may not be kind of confusing. Since the Event Store is the culmination of multiple collections, not just "streams". So it is not completely correct to say that the event store is "append-only"

Documentation/Concepts/event_store.md
```diff
@@ -2,90 +2,189 @@
...
+- You can't have [Public Streams]({{< ref "streams#public-stream" >}}) or [Aggregates]({{< ref "aggregates" >}})
+- There is a [`subscription-states`]({{< ref "#subscription-states" >}}) collection for [Subscription]({{< ref "event_horizon#subscription" >}})
+- Scoped collections have a `x-scopeID-` prefix in their names
+
+<!-- The default alert shortcode wouldn't work properly inside the tab so I copied the alert HTML here -->
+<div class="alert alert-info" role="alert">
+    <h4 class="alert-heading">Default scope</h4>
+    Technically all collections are scoped, with the default scopeID being <code>00000000-0000-0000-0000-000000000000</code>.
+    This is left out of the naming to make the event store more readable. When we talk about scoped concepts, we always refer to non-default scopes.
+</div>
```
woksin, line 164: This should maybe be out of the code tab thingy?

Documentation/Concepts/event_store.md
```diff
@@ -2,90 +2,189 @@
...
+
+#### `subscription-states`
+
+This collection keeps track of [Event Horizon Subscriptions]({{< ref "event_horizon#subscription" >}}) in a very similar way to [`stream-processor-states`]({{< ref "#stream-processor-states" >}}).
+```json
+{
+    // producers microservice and tenant
+    "Microservice": "UUID",
+    "Tenant": "UUID",
+    // consumers stream
```
woksin, line 174: This is the producer's stream I'm pretty sure. Hence the "subscription-state"

Documentation/Concepts/event_store.md
```diff
@@ -2,90 +2,189 @@
...
+        "TypeId": "UUID",
+        "TypeGeneration": "long",
+        "Public": "bool"
+    }
+}
+```
+
+### `aggregates`
+
+This collection keeps track of all [Aggregate Roots]({{< ref "aggregates#aggregates-in-dolittle" >}}) registered with the Runtime.
```
woksin, line 83: It keeps track of all Aggregates. Remember that for an AggregateRoot (a type of aggregate really) there can be multiple aggregates (with the event source as an unique identifier)

woksin (changes requested):
I got a couple of comments, other than that it looks nice

Documentation/Concepts/event_store.md
joelhoisko:
```suggestion
Events saved to the event store **cannot be changed** or **deleted**. It acts as the record of all events that have happened in the system from the beginning of time. 
```
I removed the "append-only" part.

Documentation/Concepts/event_store.md
joelhoisko: Kinda yeah, I just thought that the `x-scopeID-` part and the default scope baing an empty Guid is kinda a specific implementation detail. Or maybe the scoped part could be put up before the whole sturcure thing and then in the beginning of the strictire part I mention the scopes hmmm I'll get to it

Documentation/Concepts/event_store.md
joelhoisko: Fixd

Documentation/Concepts/event_store.md
joelhoisko: Fixed it to just aggregates. But also by definition an Aggregate Root is the only entry point to a particular aggregate, it makes no sense for an aggregate root to contain multiple aggregates within it, no?

Documentation/Concepts/event_store.md
woksin: Yes do that 👍

woksin (approved):

PR #456, opened 2021-01-14
Turbo kitchen missing logic

woksin:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199575224361229)

### DIFF
--- a/Source/Events.Processing/Streams/AbstractScopedStreamProcessor.cs
+++ b/Source/Events.Processing/Streams/AbstractScopedStreamProcessor.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Dolittle.Logging;
 using Dolittle.Resilience;
+using Dolittle.Runtime.Async;
 using Dolittle.Runtime.Events.Store.Streams;
 using Dolittle.Tenancy;
 
@@ -16,6 +18,7 @@ namespace Dolittle.Runtime.Events.Processing.Streams
     /// </summary>
     public abstract class AbstractScopedStreamProcessor
     {
+        static readonly TimeSpan EventWaiterTimeout = TimeSpan.FromMinutes(1);
         readonly IStreamDefinition _sourceStreamDefinition;
         readonly TenantId _tenantId;
         readonly IEventProcessor _processor;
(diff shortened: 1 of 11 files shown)


Source/Events.Processing/Streams/AbstractScopedStreamProcessor.cs
```diff
@@ -116,6 +119,14 @@ public Task Start(CancellationToken cancellationToken)
         /// <returns>A <see cref="Task" /> that, when resolved, returns the new <see cref="IStreamProcessorState" />.</returns>
         protected abstract Task<IStreamProcessorState> OnFailedProcessingResult(FailedProcessing failedProcessing, StreamEvent processedEvent, IStreamProcessorState currentState);
 
+        /// <summary>
+        /// Tries to get the <see cref="TimeSpan" /> for when to retry processing. Outputs the maximum value <see cref="TimeSpan" /> if there is no retry time.
+        /// </summary>
+        /// <param name="state">The current <see cref="IStreamProcessorState" />..</param>
+        /// <param name="timeToRetry">The <see cref="TimeSpan" /> for when to retry processsing a stream processor.</param>
+        /// <returns>A value indicating whether there is a retry time.</returns>
+        protected abstract bool TryGetTimeToRetry(IStreamProcessorState state, out TimeSpan timeToRetry);
```
joelhoisko, line 128:
```suggestion
        /// <summary>
        /// Tries to get the <see cref="TimeSpan" /> for when to retry processing. Outputs the maximum value <see cref="TimeSpan" /> if there is no retry time.
        /// </summary>
        /// <param name="state">The current <see cref="IStreamProcessorState" />..</param>
        /// <param name="timeToRetry">The <see cref="TimeSpan" /> for when to retry processsing a stream processor.</param>
        /// <returns>A value indicating whether there is a retry time.</returns>
        protected bool TryGetTimeToRetry(IStreamProcessorState state, out TimeSpan timeToRetry)
            =>  _timeToRetryGetter.TryGetTimespanToRetry(state as StreamProcessorState, out timeToRetry);
```

As the code for both of the implementing streamprocessors is the same here, we could just pass the `ICanGetTimeToRetryFor<T>` into the `AbstractScopedStreamProcessor`  and have it call the interface.

This would also require changes to the stream processors constructors.

joelhoisko (changes requested):

Source/Events.Processing/Streams/AbstractScopedStreamProcessor.cs
woksin: Cannot do that as that would make the AbstractScopedStreamProcessor generic, which I don't want to do since it would bubble down in lots of other parts of the code

joelhoisko: ok wont fix

joelhoisko (approved):

PR #479, opened 2021-01-19
Conservative logging

woksin:
This may be a little controversial, but I think that we are generally logging too much noise. I'm not completely sure on the "Filter" and "Event Processor" part though

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1140654762299587/1199616782132166)


joelhoisko:
Let's remember to update the first comment to adhere to our PR template when/if this gets pulled in.

PR #498, opened 2021-02-15
Features/commit events rest

woksin:
## Summary

First rough take on getting a RESTful API around the EventStore. Staying true to the gRPC protobuf contracts

### Added
- Rest endpoints around the event store in a controller
- /api/event/commit - Commits events
- /api/event/commitForAggregate - Commits aggregate events
- /api/event/fetchForAggregate - Fetch events for aggregate

### Changed
- Added a layer between the gRPC EventStore service and the Event Store which is a service that is a singleton per tenant. The new EventStore controller is also using it


woksin:
I've tried to stay as true to the gRPC protobuf contracts we have already in terms of the requests and the responses. To be able to do that I had to make a couple of ugly proxies and have some not-so-nice conversion logic going on. They are at least something that can be improved upon in the future

I have not though about the security implications, auth or which port it should be listening to. We'd need to figure those things out

woksin:
What should we do about the port and the security of these endpoints?

Source/Server/Startup.cs
```diff
@@ -40,14 +45,23 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
+                app.UseSwagger();
```
einari, line 48: I would suggest having swagger on always?  As you don't then need to set the runtime up for development and making the APIs more accessible to everyone.

einari (changes requested):
Nice..  Excited to get WebAPIs here... 

A suggestion; rename the Controllers namespace to WebAPI. 
Further more, suggesting it could be its own project as well - then we'd structurally isolate it outside of the server startup. 

The reason I say so is that we could be adding more options like GraphQL as well and its nice to keep things separate.

Source/Server/Startup.cs
woksin: Sure, that we can do.

einari:
> What should we do about the port and the security of these endpoints?

The port needs to be configurable, which I guess it is by default through the appSettings file. But we could source the configuration from environment variables - enabling us to have a port env variable. 

As for security - I suggest we come back to this question.
From our internal perspective of running it in the cluster - we don't need to expose the port right now.

woksin:
@einari So is this okay to pull in now?

einari (approved):

PR #499, opened 2021-02-17
Fix event store eventual consistency bug

woksin:
## Summary

A problem could occur when commiting multiple events because in regards to eventual consitency. This should likely solve that issue by notifying the event after the transaction has been committed. 

### Fixed

- Notify event waiter outside of the transaction

## Linked to
- https://app.asana.com/0/1199901627887462/1199949796334295/f


woksin:
I will have a second thought around how to solve this in a good way

woksin:
I think my last commit now makes it slightly more performant

freshteapot (approved):
Approved (with some words)

@woksin walked me thru the code. From a higher level, with no C# back ground it makes sense.

We agreed:
- No tests to confirm the bug
- Current tests pass
- Not 100% how to test this with Torgeirs data (but Torgeir says it works)
- @digitaldias audit tool might be possible in combination with Torgeirs data (do we need the source code) could confirm the branch that breaks and this branch that works.

Nice to learn a little about the internals.

PR #500, opened 2021-03-11
Possible fix for deadlock issue when notifying and waiting for events

woksin:
## Summary

We saw inconsistent issues while committing events in applications with event handlers that caused the runtime to deadlock when committing an event, never returning.

Inspired by this blog post https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/ we have a theory that it might be because when a TaskCompletionSource is being waited for and SetResult is called on it then all the task's async continuations will be invoked synchronously which could lead to a deadlock scenario. If that's the case then this issue can be resolved by giving the TaskCompletionSource the TaskCreationOptions.RunContinuationsAsynchronously flag when created to force the continuations to be run asynchronously. 

~~We have yet to be able to write tests that provoke this scenario. Until then we can never know whether the issue has been resolved or not, or that that was the problem to begin with.~~ I've added a single spec that failed when TaskCreationOptions.RunContinuationsAsynchronously was not given to the TaskCompletionSource and passes when it is. Thus IMO validating this fix

### Changed

- Made EventWaiter more robust
- Added TaskCreationOptions.RunContinuationsAsynchronously to all manually created TaskCompletionSources

### Added

- A spec that tries provoke the dead-lock by calling Notify and Wait in different threads and call Notify (running in separate thread from the thread calling Wait) between two Wait-calls. In this scenario the continuation of the first Wait call would be started synchronously, running on the thread that called Notify, after the Notify called TrySetResult on the TaskCompletionSource that the Wait-call awaits. And since the "Notify-thread" has the readWrite lock we get a dead-lock when the next Wait-call tries to get this lock.

### FIxed

- A wrongly formatted log message when committing events.

### DIFF
--- a/Source/Events.Store.Services/LoggerExtensions.cs
+++ b/Source/Events.Store.Services/LoggerExtensions.cs
@@ -8,13 +8,13 @@ namespace Dolittle.Runtime.Events.Store.Services
 {
     internal static class LoggerExtensions
     {
-        static readonly Action<ILogger, string, int, Exception> _eventsReceivedForCommitting = LoggerMessage
-            .Define<string, int>(
+        static readonly Action<ILogger, int, string, Exception> _eventsReceivedForCommitting = LoggerMessage
+            .Define<int, string>(
                 LogLevel.Debug,
                 new EventId(2059664795, nameof(EventsReceivedForCommitting)),
                 "{NumEvents} {EventType}events received for committing");
 
         internal static void EventsReceivedForCommitting(this ILogger logger, bool isAggregateEvents, int numEvents)
-            => _eventsReceivedForCommitting(logger, isAggregateEvents? "aggregate " : "", numEvents, null);
+            => _eventsReceivedForCommitting(logger, numEvents, isAggregateEvents? "aggregate " : "", null);
     }
 }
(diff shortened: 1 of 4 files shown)


woksin:
Should we mark this as a minor? To signify that this is a "bigger" thing? We could also deprecate and remove all versions all >= 5.3.0 and <= 5.4.1

jakhog (changes requested):
```csharp
    public class EventWaiter
    {
        readonly object _notifiedLock = new();
        readonly object _readWriteLock = new();
```

The cause of the deadlock was that there are two execution paths that captures these locks, in the opposite order. So one path would acquire `_notifiedLock` then `_readWriteLock`, while another would acquire `_readWriteLock` then `_notifiedLock`. So if one of the threads was interrupted in the middle, it would freeze the code.

The problem is alleviated with the `RunContinuationsAsynchronously`, but I think it would be nice if we weren't reliant on the scheduling to solve the deadlock issue. So my suggestion is to remove one of the locks - and just use the one.

jakhog (approved):

jakhog:
Nice! I'll pull this in as a patch trying to stay true to the SemVer scheme. And yes - we should deprecate the versions we know don't work. I'll go do it on Docker Hub - and see what we can do on the releases here.

PR #505, opened 2021-03-31
Fix enums

woksin:
## Summary

Better enum conversion

### Changed

- Change how some enums are converted between protobuf and runtime representation


jakhog (approved):
Added a few tweaks:
- Split out the extension classes to own files and added documentation
- Handled unknown types by throwing custom exceptions
- Removed some unused code to get rid of the last compiler warnings

PR #513, opened 2021-04-26
EventHandler filter validation turbo mode

jakhog:
## Summary

Adds a new Filter Validation implementation for the kind of filter (event type with eventsource partition) that validates the filter by checking just the types already committed in the event log rather than iterating through all the events.

For most registrations, the Event Handlers have not changed, which is detected by the validator and skips any further analysis. If the events in the filter has changed, the new implementation checks the types of events committed in the event log to determine if the stream would change given the new definition. Initial tests indicate that this process is a lot faster than re-filtering the streams.

### Added

- A new specialised filter validation implementation used for Event Handlers to speed up the registration process that happens during boot. This drastically reduces the CPU utilisation and duration of the initial registration process for Event Handlers.


Source/Events.Processing/Filters/ValidateFilterByComparingStreams.cs
```diff
@@ -40,7 +39,10 @@ public class ValidateFilterByComparingStreams : IValidateFilterByComparingStream
         public async Task<FilterValidationResult> Validate<TFilterDefinition>(IFilterDefinition persistedDefinition, IFilterProcessor<TFilterDefinition> filter, CancellationToken cancellationToken)
             where TFilterDefinition : IFilterDefinition
         {
-            if (persistedDefinition == default) return new FilterValidationResult();
+            if (persistedDefinition == default)
```
woksin, line 42: Nice that you made this class better as well. Since this is really similar to the new implementation could we do the same quality of life things here as well with reducing the LOC of this method by adding more methods?

Source/Events.Store.MongoDB/Streams/EventFetchers.cs
```diff
@@ -87,21 +87,23 @@ public async Task<ICanFetchEventTypesFromPartitionedStream> GetPartitionedTypeFe
             return await GetPartitionedFetcherFor(scopeId, streamDefinition, cancellationToken).ConfigureAwait(false) as ICanFetchEventTypesFromPartitionedStream;
         }
 
-        async Task<StreamFetcher<MongoDB.Events.Event>> CreateStreamFetcherForEventLog(ScopeId scopeId, CancellationToken cancellationToken) =>
-            new StreamFetcher<MongoDB.Events.Event>(
+        async Task<StreamFetcher<Events.Event>> CreateStreamFetcherForEventLog(ScopeId scopeId, CancellationToken cancellationToken) =>
```
woksin, line 90: Is the Codacy thing complaining on these qualified types? The reason I did it like this in the Events.Store.MongoDB project is because there are two separate representations of the event types; the Runtime representation in Events.Store and the MongoDB representation in Events.Store.MongoDB. And to make it clear which representation it's working on I partially qualified the types all over the place. Another way we could deal with this is to have using aliases for either the Runtime or the MongoDB types. I don't know what you think we should do?

Source/Events.Store.MongoDB/Streams/Streams.cs
```diff
@@ -20,7 +20,7 @@ public class Streams : EventStoreConnection, IStreams
         const string StreamDefinitionCollectionName = "stream-definitions";
 
         readonly ILogger _logger;
-        readonly IMongoCollection<MongoDB.Streams.StreamDefinition> _streamDefinitions;
+        readonly IMongoCollection<StreamDefinition> _streamDefinitions;
```
woksin, line 23: Same as the above comment about the qualified types

woksin (changes requested):
Nice stuff! Also nice that you added more specs and made the implementation better for validating filters by comparing streams. The only comments I have are about style really and suggestions to make it a bit easier on the eyes IMO

Source/Events.Store.MongoDB/Streams/EventFetchers.cs
jakhog: I don't think so, just my VScode passive aggressively suggesting it. But I don't really mind either way. I've reverted both files.

jakhog:
I tried to solve the last of your comments (and make Codacy more happy about duplication) by moving most of the duplicated code out to the common FilterValidators class with more sensible method names - and make a new thing for comparing the IFilterDefinition thing. I am quite happy with it. The FilterValidators implementation became quite large, but I already started going down a rabbit hole of fixing that just made a mess. So I would argue that this is enough paying back debt for one iteration. We can always come back to it.

At this point Codacy is just being stupid. It now complains about duplicated code for classes that look similar (the different validators), and for the multiple constructors it just asked me to make because it didn't like the default parameters. So seems like there is no way to please it.

It also doesn't like the out parameters. But I feel making another type just to return multiple values doesn't make it any better.

Lastly, I feel I have made tons of new tests, but it also says I have decreased the coverage. Which makes no sense to me, and I can't find an easy way of seeing where I have made things worse. As far as I can see, all the classes I have changed or implemented should be fully covered. So maybe we need to look at Codacy a bit closer at some point to see how we best use it.

woksin:
Yeah Codacy's duplication thing seems a little wonky and I would argue that we in most cases don't really care about it. And for the other rules we have not configured or customised any of them. Codacy is using SonarQube underneath for C#, and that is customisable by feeding it a config file apparently. We should spend some time at some point to configure it.

Source/Events.Processing/Filters/ValidateFilterByComparingStreams.cs
```diff
@@ -21,105 +19,85 @@ namespace Dolittle.Runtime.Events.Processing.Filters
...
-            {
-                return new FilterValidationResult($"The new stream generated from the filter will not match the old stream. {(persistedDefinition.Public ? "The previous filter is public while the new filter is not" : "The previous filter is not public while the new filter is")}");
-            }
-
-            var tryGetState = await _streamProcessorStates.TryGetFor(
-                new StreamProcessorId(filter.Scope, filter.Definition.TargetStream.Value, filter.Definition.SourceStream),
-                cancellationToken)
-                .ConfigureAwait(false);
-            if (!tryGetState.Success)
+            try
```
woksin, line 37: We can come back to this method later since it in my opinion should be made more readable by putting more things in methods :)

Source/Events.Processing/Filters/ValidateFilterForAllTenants.cs
```diff
@@ -50,21 +45,10 @@ public class ValidateFilterForAllTenants : IValidateFilterForAllTenants
...
-                    }
-                    else
-                    {
-                        _logger.NoPersistedFilterDefinition(filterProcessor.Identifier, tenantId);
-                        result.Add(tenantId, new FilterValidationResult());
-                    }
+                    _logger.ValidatingFilterForTenant(filterProcessor.Identifier, tenantId);
+                    var validationResult = await _filterValidators.Validate(filterProcessor, cancellationToken).ConfigureAwait(false);
+                    result.Add(tenantId, validationResult);
                 }).ConfigureAwait(false);
```
woksin, line 51: This can probably be done in "parallel"? I think that we have a method on _onAllTenants that does a Task.WhenAll(tasks) underneath. Maybe we could consider using it here?

woksin (commented):
Nice stuff. I still have a few comments though. They are mostly open-ended since I don't really agree with myself what we should do. Could you look at my comments again, @jakhog?

woksin (approved):

PR #516, opened 2021-05-06
Add IDetectEmbeddingLoop and EmbeddingLoopDetector

joelhoisko:
- Adds `IDetectEmbeddingLoop` and `EmbeddingLoopDetector`, which goes through the previous states in parallel using `AsParallel(IEnumerable)` and `ParallelEnumerable.Any` from LINQ.
- Adds specs for it.
    - As the class relies mostly on `CompareProjectionStates`, I don't think it makes sense to test out differently ordered properties/dictionaries etc, as those are already covered by tests by `CompareProjectionStates` specs.

### DIFF
--- a/Source/Embeddings.Processing/EmbeddingLoopDetector.cs
+++ b/Source/Embeddings.Processing/EmbeddingLoopDetector.cs
@@ -1,0 +1,35 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Dolittle.Runtime.Projections.Store.State;
+using Dolittle.Runtime.Rudimentary;
+
+namespace Dolittle.Runtime.Embeddings.Processing
+{
+    /// <inheritdoc/>
+    public class EmbeddingLoopDetector : IDetectEmdbeddingLoop
+    {
+        readonly ICompareStates _comparer = new CompareProjectionStates();
+
+        /// <inheritdoc/>
+        public Try<bool> TryCheckForProjectionStateLoop(IEnumerable<ProjectionState> previousStates, ProjectionState currentState)
+        {
+            try
+            {
+                return previousStates.All(previousState =>
+                {
+                    return !_comparer.TryCheckEquality(previousState, currentState).Result;
+                });
+            }
+            catch (Exception ex)
+            {
+                return ex;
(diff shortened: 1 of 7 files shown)


Source/Embeddings.Processing/EmbeddingLoopDetector.cs
```diff
@@ -0,0 +1,36 @@
...
+using Dolittle.Runtime.Rudimentary;
+
+namespace Dolittle.Runtime.Embeddings.Processing
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="IDetectEmbeddingLoop" />.
+    /// </summary>
+    public class EmbeddingLoopDetector : IDetectEmdbeddingLoop
+    {
+        readonly ICompareStates _comparer = new CompareProjectionStates();
```
woksin, line 17: This should be taken in as a dependency in the constructor

woksin (changes requested):
Nice. Just a few comments

Do we want to check whether there are loops in the previous states or not?

Source/Embeddings.Processing/EmbeddingLoopDetector.cs
joelhoisko: I changed it to take it as a dependency and added mocks for it in the tests.

While doing this, I also realized that the `ParallelEnumerable.Any()` does not short circuit on exceptions/`true` results, but this is a question about optimizing and if even using parallelization gives us anything here. Time will tell.

Source/Embeddings.Processing/EmbeddingLoopDetector.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    var tryResult = _comparer.TryCheckEquality(previousState, currentState);
+                    return tryResult.Success ? tryResult.Result : throw tryResult.Exception;
+                });
+            }
+            catch (Exception ex)
+            {
+                return ex;
+            }
+        }
```
woksin, line 41: Re-throwing exceptions is generally not a great idea since we lose the correct stack trace information.
I think this can be greatly simplified by doing this instead.
```suggestion
        {
            var equalityResults = previousStates.AsParallel().Select(_ => _comparer.TryCheckEquality(previousState, currentState));
            var failure = equalityResults.FirstOrDefault(_ => !_.Success);
            if (failure is not null)
            {
                return failure.Exception;
            }
            return equalityResults.Any(_ => _.Result);
        }
```

PR #518, opened 2021-05-14
Implementation and Specifications for ProjectionKey to EventSourceIdConverter.

jakhog:
## Summary

Implements the `IConvertProjectionKeysToEventSourceIds` interface.

It the 128-bit wide CityHash function, as it is suitable for non-cryptographic hashing of strings, and according to the docs should be independent of platform endianness.

### Added

- Converter for ProjectionKey -> EventSourceId


Source/Embeddings.Store/ProjectionKeyToEventSourceIdConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    public class ProjectionKeyToEventSourceIdConverter : IConvertProjectionKeysToEventSourceIds
+    {
+        readonly Encoding _encoding;
+        readonly ICityHash _hasher;
+
+        public ProjectionKeyToEventSourceIdConverter()
+        {
+            _encoding = Encoding.Unicode;
+            _hasher = CityHashFactory.Instance.Create(new CityHashConfig { HashSizeInBits = 128 });
```
woksin, line 24: This 128 number should be a variable, it is used two places. I suggest it to be a const int

woksin (changes requested):
Nice. I only have a thing about the max length magical number. Usually it's good to have those things as constants

Source/Embeddings.Store/ProjectionKeyToEventSourceIdConverter.cs
jakhog: Like what I did now? Codacy doesn't like consts apparently, it wants a static readonly property. What do we want?

woksin: I would prefer const over static readonly in this scenario as it resolves at compile-time rather than a runtime lookup. static readonly is good for initializing constant values that does not have a compile-time value. I think we'll look at tweeking some sonarqube rules at some point for codacy

woksin: Funny, I found this https://rules.sonarsource.com/csharp/tag/performance/RSPEC-3962

jakhog: How about now?

woksin: I like

PR #520, opened 2021-05-20
Remove IWaitForAggregateRootEvents and fix EmbeddingProcessor

jakhog:
## Summary

Removed the `IWaitForAggregateRootEvents` interface as it turned out to be a lot of work to implement, and it is really an optimisation. Changed the `EmbeddingProcessor` to just wait for any new events in the event log, this will cause the refreshing of embedding states to happen more often than it needs to, but we can come back to this optimisation at a later stage when we need to.

This change also required a new interface and methods extending the `StreamEventWatcher` to wait for any new event on a stream (not just after a specific position as we supported). While implementing this new behaviour, I also cleaned up the old specs and made a few more for `StreamEventWatcher`.


woksin (approved):
Nice! I have a comment on the performance implications, but this should be evaluated properly after testing

PR #521, opened 2021-05-20
Embedding service

woksin:
Implement and spec the EmbeddingsService Grpc service endpoints. Also changed the way embeddings processors are created, though this is not specced yet


Source/Embeddings.Processing/RuntimeServices.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _embeddings = embeddings;
+        }
+
+        /// <inheritdoc/>
+        public ServiceAspect Aspect => "Embeddings.Processing";
+
+        /// <inheritdoc/>
+        public IEnumerable<Service> BindServices() =>
+            new Service[]
```
woksin, line 31:
```suggestion
            new[]
```

Source/Embeddings.Processing/EmbeddingRegistrationArguments.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using Dolittle.Runtime.Artifacts;
+using Dolittle.Runtime.Embeddings.Store;
+using Dolittle.Runtime.Execution;
+using Dolittle.Runtime.Projections.Store.State;
+
+namespace Dolittle.Runtime.Embeddings.Processing
+{
+    /// <summary>
+    /// Represents the runtime representation of the embedding registartion arguments.
```
joelhoisko, line 13:
```suggestion
    /// Represents the runtime representation of the embedding registration arguments.
```

Source/Events.Store.Services.Grpc/EventStoreGrpcService.cs
```diff
@@ -56,7 +56,7 @@ public override async Task<Contracts.CommitAggregateEventsResponse> CommitForAgg
             var commitResult = await _eventStoreService.TryCommitForAggregate(
                 new UncommittedAggregateEvents(
                     eventSourceId,
-                    new Artifact(request.Events.AggregateRootId.ToGuid (), ArtifactGeneration.First),
+                    new Artifact(request.Events.AggregateRootId.ToGuid(), ArtifactGeneration.First),
```
joelhoisko, line 59: Off topic, but I know that we don't do anything with the generation right now, but shouldn't it still be set to the one from the request? Or is this just to stop from any craziness coming from the SDKs?

joelhoisko (approved):
Neato 🌯, just a quick comment.

Source/Events.Store.Services.Grpc/EventStoreGrpcService.cs
woksin: Good point. I don't remember if this was intentional or a fluke to be honest 😆

PR #524, opened 2021-06-02
Add simple callback scheduler

joelhoisko:
A simple callback scheduler using [`System.Threading.Timer`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer?view=net-5.0) with specs. The timer returns an `IDisposable`, which when disposed of, will stop the timer.

I didn't think about the singleton aspect of this, as I didn't see how that would make sense, as it's the users of the `RegisterCallback()` that are ultimately in the know about when to dispose of the callback etc.

Naming could also be thought about at some point, maybe `ScheduleCallback`?


jakhog (approved):
Nice! I'm very eager to try this out.

PR #525, opened 2021-06-03
Refactor out ping handling

jakhog:
We have already merged the interfaces so we could continue other work - sorry about that. Bat all the files have been changed here, so if you just look at the files (and ignore what GitHub says has been changed) you'll see what is new.

There are a few pieces still missing for this to work properly, that we can do later:
- Implement the `IKeepConnectionsAlive` interface (which is just a factory that will take some dependencies and pass them to the pinged connections).
- Implement and bind the `IMetricsCollector` (we can do a null implementation for now).

I purposefully did not bind the `TokenSourceDeadline` class to the interface, as I seem to recall we have issues with DI and disposables?

Also, a lot of specs for PingedConnection still missing - but it makes more sense to do that once the callback timer interface is ready.

### DIFF
--- a/Source/Services/ReverseCalls/PingedConnection.cs
+++ b/Source/Services/ReverseCalls/PingedConnection.cs
@@ -12,7 +12,7 @@ using Microsoft.Extensions.Logging;
 namespace Dolittle.Runtime.Services.ReverseCalls
 {
     /// <summary>
-    /// Represents an implementation of <see cref="IPingedReverseCallConnection{TClientMessage, TServerMessage}"/>.
+    /// Represents an implementation of <see cref="IPingedConnection{TClientMessage, TServerMessage}"/>.
     /// </summary>
     /// <typeparam name="TClientMessage">Type of the <see cref="IMessage">messages</see> that is sent from the Client to the Runtime.</typeparam>
     /// <typeparam name="TServerMessage">Type of the <see cref="IMessage">messages</see> that is sent from the Runtime to the Client.</typeparam>
@@ -20,7 +20,7 @@ namespace Dolittle.Runtime.Services.ReverseCalls
     /// <typeparam name="TConnectResponse">Type of the response that is received after the initial Connect call.</typeparam>
     /// <typeparam name="TRequest">Type of the requests sent from the Runtime to the Client.</typeparam>
     /// <typeparam name="TResponse">Type of the responses sent from the Client to the Runtime.</typeparam>
-    public class PingedReverseCallConnection<TClientMessage, TServerMessage, TConnectArguments, TConnectResponse, TRequest, TResponse> : IPingedReverseCallConnection<TClientMessage, TServerMessage>
+    public class PingedConnection<TClientMessage, TServerMessage, TConnectArguments, TConnectResponse, TRequest, TResponse> : IPingedConnection<TClientMessage, TServerMessage>
         where TClientMessage : IMessage, new()
         where TServerMessage : IMessage, new()
         where TConnectArguments : class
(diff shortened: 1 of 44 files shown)


Source/Services/ReverseCalls/TokenSourceDeadline.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed) return;
+
+            if (disposing)
+            {
+                _source.Dispose();
+            }
```
woksin, line 46: Not cancel? See previous comment about dispose pattern

Source/Services/ReverseCalls/WrappedAsyncStreamWriter.cs
```diff
@@ -68,37 +77,81 @@ public WriteOptions WriteOptions
         /// <returns>A task that will complete when the message is sent.</returns>
         public async Task WriteAsync(TServerMessage message)
         {
+            _logger.WritingMessage(_requestId, typeof(TServerMessage));
             _cancellationToken.ThrowIfCancellationRequested();
+
             if (!_writeLock.Wait(0))
```
woksin, line 83: aren't there async locks? I think there might be one in the Nito.AsyncEx packages I'm using in EventHorizon

Source/Services/ReverseCalls/WrappedAsyncStreamWriter.cs
```diff
@@ -68,37 +77,81 @@ public WriteOptions WriteOptions
...
+
                 stopwatch.Stop();
+                _metrics.DecrementPendingStreamWrites();
                 _metrics.AddToTotalStreamWriteWaitTime(stopwatch.Elapsed);
+                _logger.WritingMessageUnblockedAfter(_requestId, typeof(TServerMessage), stopwatch.Elapsed);
             }
 
-            await WriteRecordMetricsAndReleaseLock(message, false);
+            try
+            {
```
woksin, line 99: This should perhaps be moved right under the acquiring of the lock

Source/Services/ReverseCalls/WrappedAsyncStreamWriter.cs
woksin: Hmmm, I might just don't know the API well enough :P

woksin (approved):

woksin (approved):

Source/Services/ReverseCalls/PingedConnection.cs
```diff
@@ -0,0 +1,200 @@
...
+                _cancellationTokenSource.Cancel();
+                WaitForPingStarterToCompleteIgnoringExceptions();
+                MaybeStopPinging();
+                _keepAliveExpiredRegistration.Dispose();
+                _keepalive.Dispose();
+                _wrappedReader.MessageReceived -= ResetKeepaliveTokenCancellation;
+                _cancellationTokenSource.Dispose();
+                _wrappedWriter.Dispose();
+                _logger.DisposedPingedConnection(_requestId);
+            }
```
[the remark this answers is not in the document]
jakhog: We discussed this - and agreed it was correct as it is

Source/Services/ReverseCalls/TokenSourceDeadline.cs
jakhog: Same here for dispose.

Source/Services/ReverseCalls/PingedConnection.cs
```diff
@@ -0,0 +1,200 @@
...
+            _metrics.IncrementTotalKeepaliveTimeouts();
+            _logger.KeepaliveTimedOut(_requestId);
+        }
+
+        void WaitForPingStarterToCompleteIgnoringExceptions()
+        {
+            var stopwatch = Stopwatch.StartNew();
+            _logger.WaitingForPingStarterToComplete(_requestId);
+
+            _pingStarter.ContinueWith(_ => {}).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
jakhog: Maybe not clear enough by the method name, but it is a simple way to ignore exceptions and cancellations without having `GetResult()` throw an exception that we need to catch.

Source/Services/ReverseCalls/WrappedAsyncStreamWriter.cs
jakhog: In this case the `.Wait(0)` is a non-blocking attempt to take the lock. If it fails, it uses the lock asynchronously below.

Source/Services/ReverseCalls/WrappedAsyncStreamWriter.cs
jakhog: Will fix and make another PR

PR #528, opened 2021-06-03
EventHandlerService refactoring + tidying up

einari:
This tidies up the EventHandlerService by encapsulating it into an EventHandler - increases readability. 
Also added specs around the initialize and starting procedure. There could be more scenarios added, such as for the final start procedure that I didn't get time to cover right now.

Also, this PR aligns with the interfaces for PingedConnection and changes the ReverseCallDispatcher to use this.

Worth mentioning is also that I've modified the .editorconfig files and specs.props to increase developer happiness when using things like VSCode - less squigglies :)


jakhog:
Looks good - but there is a test failing after I merged in the latest branches.

PR #529, opened 2021-06-03
Client reverse calls

jakhog:
Refactoring the ReverseCallClient a little bit to hopefully make it somewhat easier to follow. It is still way to big, but changing it up much more will take a lot more time. Also changed the setup a bit to use protocol classes like we do for the dispatchers, so that we don't need so many callbacks.

Also added more logging and some metrics to the ReverseCallClient. The metrics don't do anything now - but we can fix this later.

Lastly, and most importantly - the ReverseCallClient will now accept pings while it is waiting for connect response messages. Which will make it compatible with the new ReverseCallDispatcher.


Source/Services.Clients/ReverseCallClient.cs
```diff
@@ -244,121 +159,293 @@ protected virtual void Dispose(bool disposing)
...
             if (pingInterval.TotalMilliseconds <= 0) throw new PingIntervalNotGreaterThanZero();
         }
 
-        void ThrowIfConnecting()
+        void EnsureOnlyConnectingOnce()
         {
-            if (_connecting)
+
+            if (_connecting) throw new ReverseCallClientAlreadyCalledConnect();
+            lock (this)
```
joelhoisko, line 427: I don't know much about locks in C#, but I've heard that using `lock(this)` can be [dangerous](https://stackoverflow.com/a/251668/5806412). [MS docs](https://stackoverflow.com/a/251668/5806412) also advice against it.

Source/Services.Clients/ReverseCallClient.cs
```diff
@@ -244,121 +159,293 @@ protected virtual void Dispose(bool disposing)
...
+            catch (RpcException ex) when (ex.StatusCode == StatusCode.Cancelled)
             {
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    _logger.LogDebug("Reverse Call Client was cancelled before it could respond with pong");
-                    return;
+                    _logger.DidNotReceiveConnectResponse(typeof(TConnectResponse), "Connection was closed by the client");
+                }
+                else
+                {
```
joelhoisko, line 233: Would be nice to log the ping timeouts too during connecting, like in the `Handle()`.

```suggestion
                }
                else if (keepalive.IsCancellationRequested)
                {
	                  _metrics.IncrementTotalPingTimeouts();
	                  _logger.PingTimedOut();
                }
                else
                {
```

Source/Services.Clients/ReverseCallClient.cs
```diff
@@ -244,121 +159,293 @@ protected virtual void Dispose(bool disposing)
...
+                try
+                {
+                    stopwatch.Stop();
+                    _metrics.DecrementPendingWrites();
+                    _metrics.AddToTotalWriteWaitTime(stopwatch.Elapsed);
+                    _logger.WritingMessageUnblockedAfter(typeof(TServerMessage), stopwatch.Elapsed);
+                }
+                catch
+                {
+                }
```
joelhoisko, line 394: This is a bit of a weird try catch, could it just be taken out? Or it could be changed to cover the whole pending block and then we could have logs/metrics for `FailedPendingWrite` or something.
```suggestion

                stopwatch.Stop();
                _metrics.DecrementPendingWrites();
                _metrics.AddToTotalWriteWaitTime(stopwatch.Elapsed);
                _logger.WritingMessageUnblockedAfter(typeof(TServerMessage), stopwatch.Elapsed);
```

Source/Services.Clients/ReverseCallClient.cs
```diff
@@ -108,121 +79,65 @@ public class ReverseCallClient<TClientMessage, TServerMessage, TConnectArguments
...
-                    linkedCts.CancelAfter(_pingInterval.Multiply(3));
+                    keepalive.CancelAfter(_pingInterval.Multiply(3));
                 }
             }
             catch (RpcException ex) when (ex.StatusCode == StatusCode.Cancelled)
             {
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    _logger.LogDebug("Reverse Call Client was cancelled by client while handling requests");
+                    _metrics.IncrementTotalStartedConnections();
```
joelhoisko, line 126: Wrong metrics call.
```suggestion
                    _metrics.IncrementTotalCancelledConnections();
```

Source/Services.Clients/ReverseCallClient.cs
```diff
@@ -108,121 +79,65 @@ public class ReverseCallClient<TClientMessage, TServerMessage, TConnectArguments
...
+                    _logger.ConnectionCancelledWhileHandlingRequests("client");
                     return;
                 }
 
-                if (!linkedCts.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
+                if (!keepalive.IsCancellationRequested)
                 {
-                    _logger.LogWarning("Reverse Call Client was cancelled by server while handling requests");
+                    _metrics.IncrementTotalStartedConnections();
```
joelhoisko, line 133: Wrong metrics call.
```suggestion
                    _metrics.IncrementTotalCancelledConnections();
```

joelhoisko (changes requested):
Nice, got some comments about the ReverseCallClient, most of them are about metrics/logging.

Source/Services.Clients/ReverseCallClient.cs
jakhog: I agree it looks a bit strange - but it is necessary to not leave locks locked if an error occurs. We can look into a nicer flow in the future (we have the same thing in WrappedAsyncStreamWriter).

PR #535, opened 2021-06-04
Return failure during subscribing if it's in process

joelhoisko:
Makes `Subscriptions` the mastermind behind orchestrating the subscription process. Will return a failure if a subscription with the same id is already in progress.


woksin:
Good stuff! If it's okay by you I'll build upon this :)

joelhoisko:
Yeah sure go for it, it's not an elegant solution, but I didn't really know if we wanted to do anything smarter than just telling the client that "sry we already working on this, get lost"

jakhog:
I've done a little work on this now - and some of the plans are better explained in Asana.

The missing piece is to implement the new logic in the Subscription loop method (which might require more changes). It needs to get the last received event position (I might have screwed up some dependencies here) - and then start the connection + start the stream processor - and then write the result to the TaskCompletionSource (so that the head can get back the result).

I probably have forgotten something - just let me know if the progress doesn't make any sense to anyone.

PR #537, opened 2021-06-06
Start of a Management UI 

einari:
This PR adds the starting point of a management UI. With it comes a GraphQL API surface that enables us for now to get information about event handlers. The information we're getting right now is around the position of the stream processors (Filter, EventProcessor) and also the position we're at in the source event-log of the event handler.

To achieve this there are some additions in the Runtime and to the APIs internally to the runtime to be able get enough information. Specifically, the Event Store now has on its API (`IEventStore`) the ability to get the tail event sequence number. In addition to this, the StreamProcessor now exposes the capability for retrieving the underlying abstract stream processor for a specific tenant. On this the CurrentState has now been exposed publically as read only to be able to bubble up all the information.

The frontend itself is written using React, React Router and FluentUI leveraging WebPack.

In the PR the Dockerfiles has also been updated for both the Production and the Development images that are being created.

One change this is bringing to the table is that the default Web port has changed to 80 rather than 8001. For local development, it leverages the **launchSettings.json** in the **Properties** folder giving it a default port of 5000 when running locally.

The GraphQL setup is leveraging Hot Chocolate from ChilliCream as GraphQL server.


Docker/Production/Dockerfile
```diff
@@ -22,18 +21,16 @@ WORKDIR /app/Source/Server
...
-# COPY Source/ManagementUI ./Source/ManagementUI
+COPY Source/Management.Web ./Source/Management.Web
 
-# WORKDIR /app/Source/ManagementUI
+WORKDIR /app/Source/Management.Web
 
-# RUN yarn
-# RUN yarn build
+RUN yarn
+RUN yarn build
```
jakhog, line 33: Do we want this UI thing by default in the production release - or maybe just the development+debug releases?

Haven't thought a lot about it - but I'm just wondering if there are some resource overheads or potential security problems with it on in the production version?

Source/Management.GraphQL/EventHandlers/EventHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Dolittle.Runtime.Management.GraphQL.EventHandlers
+{
+    public class EventHandler
+    {
+        public Guid Id { get; set; }
+        public Guid Scope { get; set; }
+        public int FilterPosition { get; set; }
+        public int EventProcessorPosition { get; set; }
+        public int TailEventLogSequenceNumber { get; set; }
```
jakhog, line 14: See above comments of tail-event-sequence-number. It is fine to have an int here for this and return -1 when nothing is committed.

I would maybe call the `TailEventLogSequenceNumber` the `LastCommittedEventSequenceNumber`. And since it seems like you want to report how far along the event handler has gotten - I think you want to have `LastFilteredEventStreamPosition` as well.

Source/Management.GraphQL/EventHandlers/EventHandlers.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            return _eventHandlers.All.Select(_ => new EventHandler
+            {
+                Id = _.EventProcessor,
+                Scope = _.Scope,
+                FilterPosition = (int)_.FilterStreamProcessor?.GetScopedStreamProcessorFor(tenantId).CurrentState.Position.Value,
+                EventProcessorPosition = (int)_.EventProcessorStreamProcessor?.GetScopedStreamProcessorFor(tenantId).CurrentState.Position.Value,
+                TailEventLogSequenceNumber = (int)tailEventLogSequenceNumber.Value
+            });
+        }
```
jakhog, line 46: I get that it makes sense to show things in the UI per tenant - but for the Runtime it doesn't really make sense to ask it for EventHandlers for a specific tenant. They are always for all tenants.

Wouldn't it be nicer if we just did the filtering in the view? Then you can't end up asking for things for a tenant that doesn't exist. I'm also guessing it would be nice to be able to go into a specific EventHandler and see which tenants it is failing for. And if we remove the TenantId here - we can show both things from the same fetched data.

Source/Management.GraphQL/EventHandlers/EventHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+    }
```
jakhog, line 15: There is also information available in the event store for when the stream processors last processed an event - and if it is failing (and in that case why they failed (the exception message in most cases). That would be nice to see here?

Source/Management.GraphQL/EventHandlers/EventHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+}
```
jakhog, line 16: It seems to me like this does not account for partitioned event handlers? Not sure if we need to now if the UI doesn't show failures or positions per partition.

Source/Management.GraphQL/GraphQLEndpoints.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static class GraphQLEndpoints
+    {
+        /// <summary>
+        /// Adds the GraphQL API for management to the HotChocolate GraphQL builder.
+        /// </summary>
+        /// <param name="graphQLBuilder"><see cref="IRequestExecutorBuilder"/> to add to.</param>
+        /// <returns><see cref="IRequestExecutorBuilder"/> for continuation.</returns>
+        public static IRequestExecutorBuilder AddManagementAPI(this IRequestExecutorBuilder graphQLBuilder)
+        {
+            //graphQLBuilder.AddQueryType<Query>();
```
jakhog, line 21:
```suggestion
```

Source/Management.GraphQL/Management.GraphQL.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="../../default.props"/>
+
+    <PropertyGroup>
+        <AssemblyName>Dolittle.Runtime.Management.GraphQL</AssemblyName>
```
jakhog, line 5: Not sure why this is here - but it is in all the other projects.
```suggestion
        <AssemblyName>Dolittle.Runtime.Management.GraphQL</AssemblyName>
        <RootNamespace>Dolittle.Runtime.Management.GraphQL</RootNamespace>
```

Source/Management.GraphQL/EventHandlers/EventHandler.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.Runtime.Management.GraphQL.EventHandlers
+{
+    public class EventHandler
```
jakhog, line 8: Missing some docs here.

Source/Management.GraphQL/ObjectFieldExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static IServiceProvider ServiceProvider { get; set; }
+
+        public static IObjectTypeDescriptor AddSubObject<T>(this IObjectTypeDescriptor parent, string name)
+        {
+            parent.Field(name).Type(typeof(EventHandlers.EventHandlers)).Resolve(_ => ServiceProvider.GetService<T>());
+            return parent;
+        }
+    }
+}
```
jakhog, line 20: Not sure what this does - so I'm going to ask for some docs :)

Source/Management.GraphQL/Tenancy/Tenants.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Dolittle.Runtime.Tenancy;
+
+namespace Dolittle.Runtime.Management.GraphQL.Tenancy
+{
+    public class Tenants
```
jakhog, line 11: Docs would be nice here as well :)

Source/Management.GraphQL/Tenancy/Tenants.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        readonly ITenants _tenants;
+
+        public Tenants(ITenants tenants)
+        {
+            _tenants = tenants;
+        }
+
+        public IEnumerable<Guid> All() => _tenants.All.Select(_ => _.Value);
```
jakhog, line 20: Is there no easy way to convert ConceptOf things in HotChocolate?

Source/Management.Web/App.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return (
+        <>
+            <Router>
+                <Layout />
+                <div className="content">
+                    
+                <Route exact path="/">
+                        <Dashboard/>
+                    </Route>
+                    <Route exact path="/allmetrics">
```
jakhog, line 24: If were going to have more metrics pages, they could be under a sub-path maybe? Like `/metrics/some`?
```suggestion
                    <Route exact path="/metrics">
```

Source/Management.Web/Styles/acrylic.scss
```diff
@@ -0,0 +1,45 @@
...
...2pGpeEVeCCA7b85BO3F9DzxB3cdqvBzWcmzbyMiqhzuYqtHRVG2y4x+KOlnyqla8AoWWpuBoYRxzXrfKuILl6SfiWCbjxoZJUaCBj1CjH7GIaDbc9kqBY3W/Rgjda1iqQcOJu2WW+76pZC9QG7M00dffe9hNnseupFL53r8F7YHSwJWUKP2q+k7RdsxyOB11n0xtOvnW4irMMFNV4H0uqwS5ExsmP9AxbDTc9JwgneAT5vTiUSm1E7BSflSt3bfa1tv8Di3R8n3Af7MNWzs49hmauE2wP+ttrq+AsWpFG2awvsuOqbipWHgtuvuaAE+A1Z/7gC9hesnr+7wqCwG8c5yAg3AL1fm8T9AZtp/bbJGwl1pNrE7RuOX7PeMRUERVaPpEs+yqeoSmuOlokqw49pgomjLeh7icHNlG19yjs6XXOMedYm5xH2YxpV2tc0Ro2jJfxC50ApuxGob7lMsxfTbeUv07TyYxpeLucEH1gNd4IKH2LAg5TdVhlCafZvpskfncCfx8pOhJzd76bJWeYFnFciwcYfubRc12Ip/ppIhA1/mSZ/RxjFDrJC5xifFjJpY2Xl5zXdguFqYyTR1zSp1Y9p+tktDYYSNflcxI0iyO4TPBdlRcpeqjK/piF5bklq77VSEaA+z8qmJTFzIWiitbnzR794USKBUaT0NTEsVjZqLaFVqJoPN9ODG70IPbfBHKK+/q/AWR0tJzYHRULOa4MP+W/HfGadZUbfw177G7j/OGbIs8TahLyynl4X4RinF793Oz+BU0saXtUHrVBFT/DnA3ctNPoGbs4hRIjTok8i+algT1lTHi4SxFvONKNrgQFAq2/gFnWMXgwffgYMJpiKYkmW3tTg3ZQ9Jq+f8XN+A5eeUKHWvJWJ2sgJ1Sop+wwhqFVijqWaJhwtD8MNlSBeWNNWTa5Z5kPZw5+LbVT99wqTdx29lMUH4OIG/D86ruKEauBjvH5xy6um/Sfj7ei6UUVk4AIl3MyD4MSSTOFgSwsH/QJWaQ5as7ZcmgBZkzjjU1UrQ74ci1gWBCSGHtuV1H2mhSnO3Wp/3fEV5a+4wz//6qy8JxjZsmxxy5+4w9CDNJY09T072iKG0EnOS0arEYgXqYnXcYHwjTtUNAcMelOd4xpkoqiTYICWFq0JSiPfPDQdnt+4/wuqcXY47QILbgAAAABJRU5ErkJggg==);
```
jakhog, line 43: Can you not do the $app-backround-image trick here as well?

Source/Management.Web/WebPack/devServer.js
```diff
@@ -0,0 +1,31 @@
...
+
+    return {
+        historyApiFallback: { index: basePath },
+        host: '0.0.0.0',
+        port: actualPort,
+        publicPath: basePath,
+        contentBase: process.cwd(),
+        proxy: {
+            '/api': 'http://localhost:3000',
+            '/graphql': 'http://localhost:3000',
```
jakhog, line 17: Isn't the port 5000?

Source/Management.Web/eventHandlers/EventHandlersOverview.tsx
```diff
@@ -0,0 +1,126 @@
...
+        },
+        {
+            key: 'refresh',
+            text: 'Refresh',
+            iconProps: { iconName: 'Refresh' },
+            onClick: () => getEventHandlers()
+        }
+    ];
+
+    //if (eventHandlersQueryResult.error) return <p>Error :(</p>;
```
jakhog, line 107:
```suggestion
```

Source/Management.Web/layouts/TopLevelMenu.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+    function linkClicked(item?: PivotItem, ev?: React.MouseEvent<HTMLElement>) {
+        history.push(item?.props.itemKey!);
+    }
+
+    return (
+        <div className="top-level-menu">
+            <Pivot onLinkClick={linkClicked} selectedKey={location.pathname}>
+                <PivotItem headerText="Home" itemKey="/"></PivotItem>
+                <PivotItem headerText="Metrics" itemKey="/allmetrics"></PivotItem>
```
jakhog, line 22: Goes with a previous suggestion
```suggestion
                <PivotItem headerText="Metrics" itemKey="/metrics"></PivotItem>
```

Source/Management.Web/metrics/Metrics.tsx
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { CommandBar, DetailsList, IColumn, ICommandBarItemProps, Stack } from '@fluentui/react';
+import React, { useEffect, useState } from 'react';
+import { Prometheus } from '../prometheus/Prometheus';
```
jakhog, line 6: I see you've copied a lot of code from a library. Is there no other parser of the Prometheus exposition format that would work?

Source/Management.WebAPI/Management.WebAPI.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="../../default.props"/>
+
+    <PropertyGroup>
+        <AssemblyName>Dolittle.Runtime.Management.WepAPI</AssemblyName>
```
jakhog, line 5: Same here:
```suggestion
        <AssemblyName>Dolittle.Runtime.Management.WepAPI</AssemblyName>
        <RootNamespace>Dolittle.Runtime.Management.WepAPI</RootNamespace>
```

Source/Server/Properties/launchSettings.json
```diff
@@ -0,0 +1,13 @@
+﻿{
+  "profiles": {
+    "Server": {
+      "commandName": "Project",
+      "dotnetRunMessages": "true",
+      "launchBrowser": true,
+      "applicationUrl": "http://localhost:5000",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
```
jakhog, line 9: Does this mean that we now have to explicitly set the launchSettings or the environmental variable when we deploy it in the platform?

Source/Server/ProxyServerMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    await localResponse.Body.WriteAsync(buffer.AsMemory(0, bytesRead)).ConfigureAwait(false);
+                    bytesRead = remoteStream.Read(buffer, 0, buffer.Length);
+                }
+            }
+            catch (Exception)
+            {
+            }
+        }
+    }
```
jakhog, line 68: Not sure why we need this? So maybe some docs to explain?

Source/Server/ProxyServerMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+            var localResponse = context.Response;
+            try
+            {
+                using var remoteStream = await webRequest.GetStreamAsync(url).ConfigureAwait(false);
+                var bytesRead = remoteStream.Read(buffer, 0, buffer.Length);
+
+                localResponse.Clear();
+                localResponse.ContentType = "application/text";
+
+                while (bytesRead > 0) // && localResponse.IsClientConnected)
```
jakhog, line 58:
```suggestion
                while (bytesRead > 0)
```

Source/Server/ProxyServerMiddlewareExtension.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Dolittle.Runtime.Server
+{
+    public static class ProxyServerMiddlewareExtension
+    {
+        public static IApplicationBuilder UseProxyServer(this IApplicationBuilder builder, string route, string target)
+        {
+            return builder.Use(next => new ProxyServerMiddleware(route, target, next).Invoke);
+        }
+    }
```
jakhog, line 14: Docs here would be nice as well.

Source/Server/Server.csproj
```diff
@@ -9,17 +9,19 @@
...
     <ItemGroup>
         <PackageReference Include="Dolittle.Runtime.Contracts" Version="$(ContractsVersion)" />
         <PackageReference Include="Microsoft.Extensions.Hosting" Version="$(MicrosoftExtensionsVersion)" />
         <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsVersion)" />
+        <PackageReference Include="HotChocolate.AspNetCore" Version="$(HotChocolateVersion)" />
         <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="$(AutofacExtensionsVersion)" />
         <PackageReference Include="Grpc.AspNetCore" Version="$(GrpcAspNetCoreVersion)" />
         <PackageReference Include="Grpc.AspNetCore.Web" Version="$(GrpcAspNetCoreVersion)" />
         <PackageReference Include="Swashbuckle.AspNetCore" Version="5.6.3" />
+        <PackageReference Include="GraphQL.Server.Ui.Playground" Version="5.0.1" />
```
jakhog, line 24: Versions in the versions.props file please :)

Source/Server/Startup.cs
```diff
@@ -45,20 +56,39 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+                        GraphQLEndPoint = graphQLRoute,
+                        SubscriptionsEndPoint = graphQLRoute
+                    }, graphQLPlaygroundRoute);
+                endpoints.MapGraphQL(graphQLRoute).WithOptions(new GraphQLServerOptions
+                {
+                    Tool = { Enable = false }
+                });
             });
+
+            app.UseProxyServer("/metrics", "http://localhost:9700/metrics");
```
jakhog, line 88: What is this - I don't really understand? Seems like you want to avoid CORS problems by proxying to itself onto the metrics port? It seems a bit vulnerable to hardcode the port like this here.

Can't we just serve the management UI and the metrics on the same port? That would make sense to me...

Or just let the management UI talk to the metrics directly on that port.

Source/Events.Processing/Streams/StreamProcessor.cs
```diff
@@ -68,6 +68,20 @@ public class StreamProcessor : IDisposable
...
+        /// </summary>
+        /// <param name="tenantId">Tenant to get for</param>
+        /// <returns><see cref="AbstractScopedStreamProcessor"/></returns>
+        public AbstractScopedStreamProcessor GetScopedStreamProcessorFor(TenantId tenantId)
+        {
+            if (!_streamProcessors.ContainsKey(tenantId))
+            {
+                throw new MissingScopedStreamProcessorTenant(tenantId);
+            }
+            return _streamProcessors[tenantId];
```
jakhog, line 82: I have some comments about event handlers and specific tenants later on. Would be nice if we don't have to have this exception here.

jakhog (changes requested):
Coolio. Quite a lot of comments - some of them might not be necessary to deal with now, but some small things needs to be fixed.

Docker/Production/Dockerfile
einari: We could do this for Dev only for now. The UI is one thing, but the underlying GraphQL endpoint would still be there. One way would be to feature toggle them both; have it in the image but you can basically opt in/out which would for Startup.cs mean that the GraphQL endpoints wouldn’t be added - nor the StaticFiles middleware.

Source/Events.Processing/EventHandlers/EventHandlers.cs
```diff
@@ -0,0 +1,28 @@
...
+using Dolittle.Runtime.Lifecycle;
+
+namespace Dolittle.Runtime.Events.Processing.EventHandlers
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="IEventHandlers"/>.
+    /// </summary>
+    [Singleton]
+    public class EventHandlers : IEventHandlers
+    {
```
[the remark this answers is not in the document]
einari: Completely agree - it felt really strange. I could move things around a bit to make it actually do the registration. That would make it more reliable as well.

Source/Events.Processing/EventHandlers/EventHandlers.cs
```diff
@@ -0,0 +1,28 @@
...
+        ConcurrentBag<EventHandler> _eventHandlers = new();
+
+        /// <inheritdoc/>
+        public IEnumerable<EventHandler> All => _eventHandlers.ToArray();
+
+        /// <inheritdoc/>
+        public void Register(EventHandler eventHandler) => _eventHandlers.Add(eventHandler);
+
+        /// <inheritdoc/>
+        public void Unregister(EventHandler eventHandler) => _eventHandlers = new(_eventHandlers.Except(new[] { eventHandler }));
```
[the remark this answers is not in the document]
einari: Since its a concurrent bag its working on, wouldn’t that be enough?

Source/Server/Startup.cs
einari: They have now in 4.1.1 a new AspNet support - have fixed the code by upgrading to this and taking out this. Before they didn’t have this and it spun up its own Web server on the port you dedicated, so the proxy I put in was just to be able to go through the origin without thinking about CORS.

Source/Management.GraphQL/Tenancy/Tenants.cs
einari: Yup. And I’ve done all of this in Vanir - but I didn’t feel like duplicating that logic just yet.

Source/Management.Web/App.tsx
einari: :) To be honest, this ended up being «allmetrics» because I didn’t have a better name for it since the actual metrics already exists on /metrics.  Open for namingsuggestions. The UI could provide filtering if we want to.

Source/Management.Web/Styles/acrylic.scss
einari: I can’t remember - its been a while since I did that CSS. I can give it a try.

Source/Management.Web/WebPack/devServer.js
einari: Its overridden in the WebPack.config - this whole folder is a copy/paste from Vanir.

Source/Management.Web/App.tsx
einari: The path you suggest can’t be used - without us changing the /metrics for the Prometheus endpoint.

Source/Management.Web/layouts/TopLevelMenu.tsx
einari: As before; The path you suggest can’t be used - without us changing the /metrics for the Prometheus endpoint.

Source/Management.Web/metrics/Metrics.tsx
einari: I searched around for it and couldn’t really find a good maintained library that wasn’t for Node. Not saying there isn’t one, but stumbled upon this code and just adapted it a little bit for our needs. Happy to use something completely different.

Source/Server/Properties/launchSettings.json
einari: The default behavior of Asp.Net Core is to run with http://*:80 - so we don’t need this when running.

einari:
Thanks for great in-depth review. I’ve answered some of your questions - and some with new questions :) 
But I’ve also gone ahead and fixed some of the things.

Some of it is a bit more involved, so I’ll get to that tomorrow. Especially around the EventHandlers registration - but extremely good point. Makes the code easier to reason about as well once that is done.

PR #541, opened 2021-06-09
Implement metrics collectors

jakhog:
I ended up doing all of the metrics in one go - as I thought it was nice to look at the names together to make sure they are consistent.

I've tried to base the naming of the Prometheus best practices naming conventions https://prometheus.io/docs/practices/naming/.

All our metrics start with `dolittle_`, then either `system_`, `customer_` or `shared_` (indicating whether the metric is meant for us, the customer, or both). Then the namespace (replace `.` with `_`), lastly the actual metric name and the unit following the Prometheus guide.

I found some inconsistencies with the method names in the interfaces, but didn't fix them. I rather tried to make the metric names consistent (as those are what we will see outside the Runtime).

Also I found a few small bugs related to metrics that I fixed.

To try it out, start the Runtime (and a Head) - then look at the metrics on http://localhost:9700/metrics. You get the latest metrics every time you refresh the page.


woksin (approved):
LGTM

PR #546, opened 2021-06-16
Add new specs for ReverseCallClient

joelhoisko:
- More specs for ReverseCallClient
- Removes some extraneous calls to the `_serverToClient.Current` to make specs simpler.


woksin (approved):
This looks good

PR #549, opened 2021-06-22
Specs and implementations for validating and persisting embedding definition

woksin:
This was a lot since I packed both validating and persisting definitions AND using it in the EmbeddingService... So sorry :D 

I also had to fix a mistake where the embedding definition was defined to have event selectors instead of just artifacts. This trickled down into the Embedding MongoDB store.


Source/Embeddings.Processing/CompareProjectionDefinitionsForAllTenants.cs
```diff
@@ -0,0 +1,111 @@
...
+using Dolittle.Runtime.Tenancy;
+using System;
+
+namespace Dolittle.Runtime.Embeddings.Processing
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="ICompareEmbeddingDefinitionsForAllTenants" />.
+    /// </summary>
+    [Singleton]
+    public class CompareEmbeddingDefinitionsForAllTenants : ICompareEmbeddingDefinitionsForAllTenants
```
joelhoisko, line 21: The filename should be `CompareEmbeddingDefinitionsForAllTenants.cs`

Source/Embeddings.Processing/CompareProjectionDefinitionsForAllTenants.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            var results = new Dictionary<TenantId, EmbeddingDefinitionComparisonResult>();
+            await _onTenants.PerformAsync(async tenant =>
+            {
+                var definitions = _getDefinitions();
+                var tryGetDefinition = await definitions.TryGet(definition.Embedding, token).ConfigureAwait(false);
+                var comparisonResult = tryGetDefinition.Success switch
+                {
+                    true => DefinitionsAreEqual(definition, tryGetDefinition),
+                    false => EmbeddingDefinitionComparisonResult.Equal
```
joelhoisko, line 52: So if the definition isn't persisted, then `tryGetDefinition.Success == false` and we will return saying that it was `EmbeddingDefinitionComparisonResult.Equal`? This feels a bit wrong, but I'm not sure if there is a big enough reason to have a third case, like `EmbeddingDefinitionComparisonResult.NotInitialized` or something. Now it just feels a bit wonky that a method `DiffersFromPersisted` returns equal, when the passed definition differs from persisted (which doesn't exist).

Source/Embeddings.Processing/EmbeddingsService.cs
```diff
@@ -191,5 +215,46 @@ bool TryGetRegisteredEmbeddingProcessorForTenant(TenantId tenant, EmbeddingId em
...
+            if (!tenantsAndComparisonResult.Values.Any(_ => !_.Succeeded))
+            {
+                return false;
+            }
+            var unsuccessfulComparisons = tenantsAndComparisonResult
+                .Where(_ => !_.Value.Succeeded)
+                .Select(_ =>
+                {
+                    return (_.Key, _.Value);
+                });
```
joelhoisko, line 241:
```suggestion
                .Select(_ => (_.Key, _.Value));
```
Is there something magical with tuples so that we can't just do this instead?

Source/Projections.Store.MongoDB/ProjectionsResourceTypeRepresentation.cs
```diff
@@ -16,9 +16,8 @@ public class ProjectionResourceTypeRepresentation : IRepresentAResourceType
     {
         static readonly IDictionary<Type, Type> _bindings = new Dictionary<Type, Type>
         {
-            { typeof(IProjections), typeof(Projections) },
+            { typeof(IProjectionStates), typeof(ProjectionStates) },
```
joelhoisko, line 19: Didn't we have an automatic DI convention for stuff like `IClass` => `Class`?

Source/Server/.dolittle/bounded-context.json
```diff
@@ -3,11 +3,15 @@
...
+        "eventStore": {
+            "production": "MongoDB",
+            "development": "MongoDB"
+        },
         "projections": {
             "production": "MongoDB",
             "development": "MongoDB"
         },
-        "eventStore": {
+        "embeddings": {
```
joelhoisko, line 14: Is this file used by the new stuff (projections, embeddings) still? I know that it's required for the Runtime to boot, but I thought that it was only used by some old stuff.

joelhoisko (changes requested):
Nice, lots of specs 👍 Some comments and thoughts on things, mostly just questions.

Source/Embeddings.Processing/CompareProjectionDefinitionsForAllTenants.cs
woksin: Yeah I know. It's how the projections stuff does it now. There is no grand reason behind this. I think that we could fix this by changing the interface name. I think that we should note this and fix this together with doing the same thing for projections

Source/Embeddings.Processing/EmbeddingsService.cs
woksin: Yes, it's something magical :)

Source/Server/.dolittle/bounded-context.json
woksin: Yes it is used by the resource system for hooking up services with the correct implementation types (which we don't really use since we just have MongoDB...) But it need to stay for the time being

Source/Projections.Store.MongoDB/ProjectionsResourceTypeRepresentation.cs
woksin: Yes, but this is for the resource system that was initially built to allow us in having multiple implementations for resources (like sql event store or whatever). We don't really use those capabilities since we only have a MongoDB implementation.

joelhoisko (approved):

PR #551, opened 2021-06-23
Set correct event source

woksin:
The wrong event source id for uncommitted events coming from embedding was being set. The event source id to use should be the one converted from the projection key.


joelhoisko (approved):
👍👍

PR #589, opened 2021-11-09
Legolas - Tenants and Resources 

woksin:
## Summary

Implements Contracts 6.3.0, which adds support for getting the configured Tenants, and connection string for a MongoDB resource per tenant. Note: this MongoDB connection string is configured through the "readModels" key in `resources.json` to be backwards compatible with the platform configuration.

### Added

- TenantsService: Getting all tenants through gRPC
- ResourcesService: Getting a specific resource through gRPC. Currently only the MongoDB is supported through the readModels resource configuration


Source/Resources/MongoDB/Configuration.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// Gets or sets the database name.
+        /// </summary>
+        public string Database { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value indicating whether or not to use SSL.
+        /// </summary>
+        /// <remarks>This is not used for anything.</remarks>>
+        public bool UseSSL { get; set; }
```
jakhog, line 25: Check if this is used or not

Source/Resources/MongoDB/ConnectionDetails.cs
```diff
@@ -0,0 +1,14 @@
...
+using MongoDB.Driver;
+
+namespace Dolittle.Runtime.Resources.MongoDB
+{
+    /// <summary>
+    /// Represents the connection details for a MongoDB resource.
+    /// </summary>
+    /// <param name="ConnectionString">The connection string.</param>
+    /// <param name="DatabaseName">The MongoDB resource database name.</param>
+    public record ConnectionDetails(MongoUrl ConnectionString, string DatabaseName);
```
jakhog, line 13: Doesn't the MongoUrl already contain the DatabaseName?

Source/Resources/MongoDB/IResource.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Defines a system that knows about the MongoDb resource.
+    /// </summary>
+    public interface IResource
+    {
+        /// <summary>
+        /// Gets the <see cref="MongoUrl"/> for the MongoDB resource.
+        /// </summary>
+        /// <returns>The <see cref="MongoUrl"/> connection string.</returns>
+        MongoUrl GetConnectionString();
```
jakhog, line 17: Should we be using MongoUrl or ConnectionString here, are they actually the same?

Source/Resources/MongoDB/Resource.cs
```diff
@@ -0,0 +1,37 @@
...
+    [SingletonPerTenant]
+    public class Resource : IResource
+    {
+        readonly IConfigurationFor<Configuration> _configuration;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Resource"/> class. 
+        /// </summary>
+        /// <param name="configuration">The <see cref="IConfigurationFor{T}"/> <see cref="Configuration"/>.</param>
+        public Resource(IConfigurationFor<Configuration> configuration)
```
jakhog, line 22: So we don't need to access the resource type configuration in a special way?

Source/Resources/MongoDB/ResourceType.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Represents a <see cref="IAmAResourceType">resource type</see> for the Mongo Db resource.
+    /// </summary>
+    /// <remarks>The name is currently readModels to support for legacy cases.</remarks>
+    public class ResourceType : IAmAResourceType
+    {
+        /// <summary>
+        /// The <see cref="ResourceTypes.ResourceType"/> name.
+        /// </summary>
+        public static ResourceTypes.ResourceType ResourceTypeName => "readModels";
```
jakhog, line 19: Let's make a remark that this is called "readModels" and not something with MongoDB to be backwards compatible with the old style of resource configuration and use the old databases. Also - let's not this in the release notes so we don't get confused in the future.

Source/Resources/MongoDB/ResourceTypeRepresentation.cs
```diff
@@ -0,0 +1,30 @@
...
+        static readonly IDictionary<Type, Type> _bindings = new Dictionary<Type, Type>
+        {
+            { typeof(IResource), typeof(Resource) }
+        };
+
+        /// <inheritdoc/>
+        public ResourceTypes.ResourceType Type => ResourceType.ResourceTypeName;
+
+        /// <inheritdoc/>
+        public ResourceTypeImplementation ImplementationName => "n/a";
```
jakhog, line 22: Same here, let's make a not of this strange behaviour (and verify that it actually doesn't matter).

Source/Resources/MongoDB/Service.cs
```diff
@@ -0,0 +1,51 @@
...
+                return new GetMongoDbResponse(new GetMongoDbResponse
+                {
+                    ConnectionString = _getMongoDbResource().GetConnectionString().ToString()
+                });
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex, "Failed to get MongoDb resource for tenant {Tenant}", executionContext.Tenant.Value);
+                return new GetMongoDbResponse {Failure = new Failure(ex.Message) };
+            }
```
jakhog, line 48: Change loggers to the fancy-fast way of doing it.

Source/Server/.dolittle/bounded-context.json
```diff
@@ -14,10 +14,14 @@
         "embeddings": {
             "production": "MongoDB",
             "development": "MongoDB"
+        },
+        "readModels": {
+            "development": "n/a",
+            "development": "n/a"
```
jakhog, line 20: Ahh - this might be why it works. Change this to look more like it does in Production.

jakhog: Also, as codacy says - one of these should be production.

Source/Tenancy/LoggerExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// Represents a extensions for <see cref="ILogger" />.
+    /// </summary>
+    static class LoggerExtensions
+    {
+        static readonly Action<ILogger, Exception> _failure = LoggerMessage
+            .Define(
+                LogLevel.Warning,
+                new EventId(1231141, nameof(Failure)),
+                "An error occurred");
```
jakhog, line 18: Seems like we can make a little more specific message for what's happening here.

Source/Tenancy/Tenancy.csproj
```diff
@@ -1,16 +1,24 @@
...
+        <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="../ApplicationModel/ApplicationModel.csproj" />
+        <ProjectReference Include="../Configuration/Configuration.csproj" />
+        <ProjectReference Include="../Execution/Execution.csproj" />
+        <ProjectReference Include="../Types/Types.csproj" />
+        <ProjectReference Include="../Serialization.Json/Serialization.Json.csproj" />
+        <ProjectReference Include="..\Services\Services.csproj" />
```
jakhog, line 21: Fix slash direction.

versions.props
```diff
@@ -3,7 +3,7 @@
         <AutofacVersion>5.0.0</AutofacVersion>
         <AutofacExtensionsVersion>6.0.0</AutofacExtensionsVersion>
         <ConsoleTablesVersion>2.4.2</ConsoleTablesVersion>
-        <ContractsVersion>6.2.0</ContractsVersion>
+        <ContractsVersion>6.3.0-legolas.3</ContractsVersion>
```
jakhog, line 6: Upgrade to latest.

jakhog (commented):
Added some comments for myself to look closer at and try out. Will probably fix or ignore most of these myself :)

Source/Resources/MongoDB/ConnectionDetails.cs
woksin: It does. I think that record isn't used anymore?

Source/Resources/MongoDB/IResource.cs
woksin: The MongoUrl is essentially the connection string

Source/Resources/MongoDB/Resource.cs
woksin: I was also thinking that, but I think that is handled bæin the background when getting the IConfogurationFor

Source/Tenancy/Tenancy.csproj
woksin: Rider does that automatically. Would be nice if we could edit that in Rider

versions.props
jakhog: Done

Source/Tenancy/Tenancy.csproj
jakhog: Done

Source/Server/.dolittle/bounded-context.json
jakhog: Done

Source/Resources/MongoDB/ResourceTypeRepresentation.cs
jakhog: Done.

Source/Resources/MongoDB/Resource.cs
jakhog: Yep - this is the same way we've been doing it for the EventStoreConfiguration :)

Source/Resources/MongoDB/ResourceType.cs
jakhog: Done

Source/Resources/MongoDB/Configuration.cs
jakhog: It is used, so deleted the remark.

Source/Resources/MongoDB/ConnectionDetails.cs
jakhog: File wasn't used anymore, so I got rid of it.

Source/Resources/MongoDB/IResource.cs
jakhog: Turns out MongoDB uses the MongoUrl under the hood: https://github.com/mongodb/mongo-csharp-driver/blob/30eded69a28ef7be8a01fdc7952dfa66a2fac8d8/src/MongoDB.Driver/MongoClientSettings.cs#L830 so this seems to be the correct thing to use.

Source/Tenancy/LoggerExtensions.cs
jakhog: Done

Source/Resources/MongoDB/Service.cs
jakhog: Fixed.

PR #593, opened 2021-11-15
Minor fixes and suggestions

jakhog:
- Uses new Contracts (with call context and fixed caps for MongoDB)
- Attempted to rename IService and IResource in MongoDB and slightly changed structure (this is very up for discussion)
- Made log messages the same and used the fancy-fast way of logging for both new services
- Some other minor touchups of docs ++


woksin (approved):
LGTM. I like these names more

PR #598, opened 2021-12-02
Update to net6.0

woksin:
## Summary

Updates the contracts to Gimli prerelease and updates the Runtime to .net 6.

### DIFF
--- a/Source/CLI/CLI.csproj
+++ b/Source/CLI/CLI.csproj
@@ -5,7 +5,7 @@
         <AssemblyName>Dolittle.Runtime.CLI</AssemblyName>
         <RootNamespace>Dolittle.Runtime.CLI</RootNamespace>
         <OutputType>Exe</OutputType>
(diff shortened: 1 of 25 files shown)


Source/CLI/CLI.csproj
```diff
@@ -5,7 +5,7 @@
         <AssemblyName>Dolittle.Runtime.CLI</AssemblyName>
         <RootNamespace>Dolittle.Runtime.CLI</RootNamespace>
         <OutputType>Exe</OutputType>
-        <TargetFramework>net5.0</TargetFramework>
+        <TargetFramework>net6.0</TargetFramework>
```
jakhog, line 8: Should this maybe be multi-target to support other runtimes for users?

jakhog (commented):
Looks nice :) But I'm wondering if the CLI tool should be multi-targeted?

PR #610, opened 2022-01-21
Implement service

jakhog:
## Summary

Implement the `GetAllInBatches` method so that it splits the retrieved states into batches of ~2MB. It does protect against the case where a single state is bigger than that, as there is really nothing we can do at that point (we could throw an exception in the Runtime, but we could be lucky and the client accepts it). Also added some specs that check that it splits as expected, they do not verify the content of the messages right now.


jakhog:
I can test this locally when I think about it...

PR #612, opened 2022-01-22
Batch cleaner

woksin:
Attempt to make the projection batching logic look cleaner


woksin:
@jakhog I've reviewed your #609 PR and it looks good. The only thing I wanted to try out was to make the batching logic a bit less "branched", reduce the cognitive load a bit. In my opinion this looks a bit nicer, but that's just my opinion. What do you think?

woksin:
I think that it should behave the same as your version in terms of using the async enumerator

PR #621, opened 2022-02-03
Renaming and new contracts structure

jakhog:
## Summary

- Updated Contracts to latest prerelease
- Implemented new conversion structure received from Contracts -> Runtime -> MongoDB
- Introduce and implement IPropertyRenamer that takes care of renaming properties in read models


woksin (approved):
Nice loooots of specs

PR #625, opened 2022-02-09
Make the pippin prerelease

woksin:
## Summary

Summary of the PR here. The GitHub release description is created from this comment so keep it nice and descriptive.

Remember to remove sections that you don't need or use.

### Added

- Describe the added features

### Changed

- Describe the outwards facing code change

### Fixed

- Describe the fix and the bug

### Removed

- Describe what was removed and why

### Security

- Describe the security issue and the fix

### Deprecated

- Describe the part of the code being deprecated and why

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 name: Release
 
 env:
-  PRERELEASE_BRANCHES: sam # Comma separated list of prerelease branch names. 'alpha,rc, ...'
+  PRERELEASE_BRANCHES: pippin # Comma separated list of prerelease branch names. 'alpha,rc, ...'
 
 on:
   pull_request:


Source/Events.Processing/Projections/IValidateOccurredFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+/// Defines a validator for <see cref="OccurredFormat"/>.
+/// </summary>
+public interface IValidateOccurredFormat
+{
+    /// <summary>
+    /// Checks whether the given <see cref="OccurredFormat"/> is valid.
+    /// </summary>
+    /// <param name="format">The <see cref="OccurredFormat"/>to check</param>
+    /// <param name="errorMessage"></param>
+    /// <returns></returns>
```
jakhog, line 18:
```suggestion
    /// <param name="errorMessage">A message describing why the format is invalid if validation fails, null otherwise.</param>
    /// <returns>True if the format is valid, false if not.</returns>
```

Source/Events.Processing/Projections/IValidateOccurredFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+    bool IsValid(OccurredFormat format, out string errorMessage);
```
jakhog, line 19: Any reason why this thing doesn't out the `Exception` directly?

Source/Events.Processing/Projections/IValidateOccurredFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Dolittle.Runtime.Events.Processing.Projections;
+
+/// <summary>
+/// Defines a validator for <see cref="OccurredFormat"/>.
+/// </summary>
+public interface IValidateOccurredFormat
+{
+    /// <summary>
+    /// Checks whether the given <see cref="OccurredFormat"/> is valid.
```
jakhog, line 14:
```suggestion
    /// Checks whether the given <see cref="OccurredFormat"/> is valid, meaning it can be used to format <see cref="DateTimeOffset"/> to a string.
```

Source/Events.Processing/Projections/OccurredFormatCannotBeNullOrEmpty.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// The exception that gets thrown when validating occurred format and it's null or empty.
+/// </summary>
+public class OccurredFormatCannotBeNullOrEmpty : Exception
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OccurredFormatCannotBeNullOrEmpty"/> class.
+    /// </summary>
+    public OccurredFormatCannotBeNullOrEmpty()
+        : base($"Occurred format cannot be null or empty")
```
jakhog, line 17:
```suggestion
        : base("Occurred format cannot be null or empty")
```

Source/Events.Processing/Projections/ProjectionKeys.cs
```diff
@@ -17,61 +20,89 @@ namespace Dolittle.Runtime.Events.Processing.Projections;
...
-        if (eventSelector == null)
+        try
         {
+            return eventSelector != null && TryGetKey(eventSelector, @event, partition, out key);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "Failed to get projection key for key selector type {KeySelectorType}", eventSelector.KeySelectorType);
+            key = null;
             return false;
```
jakhog, line 48: This could have happened just in the OccuredIsKey method, since all the other methods don't throw.

Source/Events.Processing/Projections/ProjectionsProtocol.cs
```diff
@@ -87,6 +95,13 @@ public ConnectArgumentsValidationResult ValidateConnectArguments(ProjectionRegis
                 return ConnectArgumentsValidationResult.Failed($"Event {eventType.Key.Value} was specified more than once");
             }
         }
+        foreach (var (eventType, occurredSelector) in arguments.ProjectionDefinition.Events.Where(_ => _. KeySelectorType == ProjectEventKeySelectorType.Occurred).ToDictionary(_ => _.EventType, _ => _))
+        {
+            if (!_occurredFormatValidator.IsValid(occurredSelector.OccurredFormat, out var errorMessage))
+            {
+                return ConnectArgumentsValidationResult.Failed($"Event {eventType} has key from event occurred selector with an invalid occurred format {occurredSelector.OccurredFormat}. {errorMessage}");
+            }
+        }
```
jakhog, line 104: Could we move this to its own method?

Source/Events.Processing/Projections/ValidateOccurredFormat.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+            errorMessage = null;
+            _ = DateTimeOffset.UtcNow.ToString(format, CultureInfo.InvariantCulture);
+            return true;
+        }
+        catch (Exception ex)
+        {
+            errorMessage = ex.Message;
+            return false;
+        }
```
jakhog, line 31: Yeah, this seems to me like it would be nicer if we just out the exception :) No need to throw and catch.

Source/Projections.Store/Definition/OccurredFormat.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.Runtime.Rudimentary;
+
+namespace Dolittle.Runtime.Projections.Store.Definition;
+
+/// <summary>
+/// Represents the format used to represent the date time mapping of when events occurred to projections.
+/// </summary>
+/// <param name="Value">The key selector expression as a string.</param>
```
jakhog, line 11:
```suggestion
/// <param name="Value">The occurred format as a string.</param>
```

Source/Projections.Store/Definition/OccurredFormat.cs
```diff
@@ -0,0 +1,19 @@
...
+public record OccurredFormat(string Value) : ConceptAs<string>(Value)
+{
+    /// <summary>
+    /// Implicit operator from string.
+    /// </summary>
+    /// <param name="key">The occurred format.</param>
+    public static implicit operator OccurredFormat(string key) => new(key);
```
jakhog, line 18: In stead of the validator, we could also do it here in the concept. There could be an IsValid method that checks, that is used to check the registration arguments - and then in the constructor here we could throw if IsValid returns false. That way, it shouldn't be possible to sneakily create an invalid format anywhere.

Source/Projections.Store/Definition/ProjectEventKeySelectorType.cs
```diff
@@ -21,5 +21,15 @@ public enum ProjectEventKeySelectorType : ushort
...
+        
+    /// <summary>
+    /// The type when a constant, static, string is the key.
+    /// </summary>
+    Static,
+    
+    /// <summary>
+    /// The type when the date time is the key.
+    /// </summary>
+    Occurred
```
jakhog, line 34: Would it be more consistent to call this?
```suggestion
    EventOccurred
```

jakhog (changes requested):
Some small mistakes, and some suggestions

Source/Events.Processing/Projections/IValidateOccurredFormat.cs
woksin: Whoops, thanks :)

Source/Events.Processing/Projections/ProjectionsProtocol.cs
woksin: Yes

Source/Projections.Store/Definition/OccurredFormat.cs
woksin: Problem with that is that it's "allowed" that it's empty or null when the KeySelectorType is not Occurred, see how the KeySelector instances are created with empty strings for the fields that are not used

Source/Events.Processing/Projections/IValidateOccurredFormat.cs
woksin: Done

Source/Events.Processing/Projections/ProjectionsProtocol.cs
woksin: Done

woksin:
This has now be tested with both SDKs and should be ready to merge

jakhog (approved):

PR #644, opened 2022-03-04
Fix development dockerfile

woksin:
## Summary

Multi target build of development Dockerfile


woksin:
Retrigger build when contracts has been released

PR #660, opened 2022-04-05
Make it build again

woksin:
## Summary

Fix projects after breaking change to event store

### DIFF
--- a/Source/Embeddings.Processing/EmbeddingProcessor.cs
+++ b/Source/Embeddings.Processing/EmbeddingProcessor.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Dolittle.Runtime.DependencyInversion;
@@ -172,7 +173,8 @@ public class EmbeddingProcessor : IEmbeddingProcessor
                 return new FailedToCommitEmbeddingEvents(response.Failure.ToFailure());
             }
 
-            await replaceOrRemoveEmbedding(response.Events.ToCommittedEvents()[^1].AggregateRootVersion + 1).ConfigureAwait(false);
+            var committedEvents = response.Events.ToCommittedEvents();
+            await replaceOrRemoveEmbedding(committedEvents.Last().AggregateRootVersion + 1).ConfigureAwait(false);
             return Try.Succeeded();
         }
         catch (Exception ex)
--- a/Source/Embeddings.Processing/EmbeddingProcessor.cs
+++ b/Source/Embeddings.Processing/EmbeddingProcessor.cs
@@ -3,18 +3,23 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Dolittle.Runtime.DependencyInversion;
 using Dolittle.Runtime.Domain.Tenancy;
 using Dolittle.Runtime.Embeddings.Store;
+using Dolittle.Runtime.Events.Contracts;
 using Dolittle.Runtime.Events.Store;
 using Dolittle.Runtime.Events.Store.Streams;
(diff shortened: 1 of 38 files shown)


Source/Embeddings.Processing/EmbeddingProcessor.cs
```diff
@@ -163,8 +167,14 @@ async Task<Try> Loop()
...
-            var committedEvents = await _eventStore.CommitAggregateEvents(uncommittedEvents.Result, executionContext, cancellationToken).ConfigureAwait(false);
-            await replaceOrRemoveEmbedding(committedEvents[^1].AggregateRootVersion + 1).ConfigureAwait(false);
+            var response = await _eventStore.CommitAggregateEvents(uncommittedEvents.Result.ToCommitRequest(executionContext), cancellationToken).ConfigureAwait(false);
+            if (response.Failure != default)
+            {
+                return new FailedToCommitEmbeddingEvents(response.Failure.ToFailure());
+            }
+
+            var committedEvents = response.Events.ToCommittedEvents();
+            await replaceOrRemoveEmbedding(committedEvents.Last().AggregateRootVersion + 1).ConfigureAwait(false);
```
woksin, line 177: Does this make sense, @jakhog ?

Source/Events.Store/CommittedAggregateEventsExtensions.cs
```diff
@@ -29,10 +29,16 @@ public static Contracts.CommittedAggregateEvents ToProtobuf(this CommittedAggreg
...
+    /// <summary>
+    /// Converts the <see cref="Contracts.CommittedAggregateEvents"/> to <see cref="CommittedAggregateEvents"/>.
+    /// </summary>
+    /// <param name="committedAggregateEvents">The committed events.</param>
+    /// <returns>The converted <see cref="CommittedAggregateEvents"/>.</returns>
     public static CommittedAggregateEvents ToCommittedEvents(this Contracts.CommittedAggregateEvents committedAggregateEvents)
     {
-        var version = committedAggregateEvents.AggregateRootVersion - (ulong)committedAggregateEvents.Events.Count;
+        var version = committedAggregateEvents.AggregateRootVersion + 1 - (ulong)committedAggregateEvents.Events.Count;
         return new CommittedAggregateEvents(
```
woksin, line 42: I see this method is only used one place currently (in the Embeddings), but does it make sense? I'll need to make some specs for this :) @jakhog

Source/Embeddings.Processing/EmbeddingProcessor.cs
jakhog: This looks to me like the same behaviour as before - so from a quick look, it looks correct :) Aren't there specs that check this?

Source/Events.Store/CommittedAggregateEventsExtensions.cs
jakhog: I'm not sure why we need to make the +1 change? And shouldn't it be possible to reuse the conversion that is used in FetchForAggregate? Should be the same thing right?

Source/Embeddings.Processing/EmbeddingProcessor.cs
woksin: There are a few :)

Source/Events.Store/CommittedAggregateEventsExtensions.cs
woksin: Maybe I'm just stupid here. Let me have a look :)

woksin: Naaah cannot use it since the stored event(in FetchForAggregate) has the aggregate root version on it. The contract version of committed aggregate event does not. I'm pretty sure the the +1 is correct here (the specs says so :P) I will make specs for it though

woksin: I've written specs that should cover this method

PR #690, opened 2022-05-03
Hinder event horizon from polling the event store

woksin:
## Summary

Fixes an issue introduced many versions ago related to event horizon producer non-stop polling the event store for an event that isn't there

### DIFF
--- a/Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
+++ b/Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
@@ -47,6 +47,10 @@ public class EventsFromEventHorizonFetcher : ICanFetchEventsFromStream, IStreamE
         }
     }
 
+    /// <inheritdoc />
+    public async Task<Try<StreamPosition>> GetNextStreamPosition(CancellationToken cancellationToken)
+        => new NotImplementedException();
+
     /// <inheritdoc/>
     public void NotifyForEvent(ScopeId scope, StreamId stream, StreamPosition position)
     {
--- a/Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
+++ b/Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
@@ -49,7 +49,7 @@ public class EventsFromEventHorizonFetcher : ICanFetchEventsFromStream, IStreamE
 
     /// <inheritdoc />
     public async Task<Try<StreamPosition>> GetNextStreamPosition(CancellationToken cancellationToken)
-        => new NotImplementedException();
+        => new NotImplementedException("GetNextStreamPosition should never be used on this specific fetcher");
 
     /// <inheritdoc/>
     public void NotifyForEvent(ScopeId scope, StreamId stream, StreamPosition position)
--- a/Source/EventHorizon/Producer/EventHorizon.cs
+++ b/Source/EventHorizon/Producer/EventHorizon.cs
@@ -149,9 +149,14 @@ public class EventHorizon : IDisposable
                         _cancellationTokenSource.Token).ConfigureAwait(false);
                     if (!tryGetStreamEvent.Success)
                     {
+                        var nextPosition = await publicEvents.GetNextStreamPosition(_cancellationTokenSource.Token).ConfigureAwait(false);
+                        if (!nextPosition.Success)
+                        {
+                            throw nextPosition.Exception;
+                        }
                         await eventWaiter.WaitForEvent(
                             Id.PublicStream,
-                            CurrentPosition,
+                            nextPosition,
                             TimeSpan.FromMinutes(1),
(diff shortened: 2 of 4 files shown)


Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
```diff
@@ -47,6 +47,10 @@ public async Task<Try<IEnumerable<StreamEvent>>> Fetch(StreamPosition streamPosi
         }
     }
 
+    /// <inheritdoc />
+    public async Task<Try<StreamPosition>> GetNextStreamPosition(CancellationToken cancellationToken)
+        => new NotImplementedException();
```
jakhog, line 52: I don't know if there is a "reason" kind of option here? But maybe it is useful for ourselves to remember why it is not implemented?

Source/EventHorizon/Producer/EventHorizon.cs
```diff
@@ -149,9 +149,14 @@ async Task ProcessEventsThroughEventHorizon()
                         _cancellationTokenSource.Token).ConfigureAwait(false);
                     if (!tryGetStreamEvent.Success)
                     {
+                        var nextPosition = await publicEvents.GetNextStreamPosition(_cancellationTokenSource.Token).ConfigureAwait(false);
```
jakhog, line 152: This does what it should - if anything maybe the name of the variable should reflect that it is for all partitions. Since everything else in this file is scoped to a specific partition?

Source/EventHorizon/Producer/EventHorizon.cs
```diff
@@ -149,9 +149,14 @@ async Task ProcessEventsThroughEventHorizon()
...
+                        if (!nextPosition.Success)
+                        {
+                            throw nextPosition.Exception;
```
jakhog, line 155: We could just wait a little bit and continue the loop here - but maybe this exception is caught and retried elsewhere?

Source/Events.Store.MongoDB/Streams/StreamFetcher.cs
```diff
@@ -111,6 +111,20 @@ public async Task<Try<IEnumerable<StreamEvent>>> Fetch(StreamPosition streamPosi
         }
     }
 
+    /// <inheritdoc />
+    public async Task<Try<StreamPosition>> GetNextStreamPosition(CancellationToken cancellationToken)
+    {
+        try
+        {
+            var numEvents = await _collection.CountDocumentsAsync(_filter.Empty, cancellationToken: cancellationToken).ConfigureAwait(false);
```
jakhog, line 119: I don't know, but it might be possible that getting the last document sorted ascending is faster? But we could just leave this for optimisation later...

Source/Events.Store/Streams/ICanFetchEventsFromStream.cs
```diff
@@ -20,4 +20,11 @@ public interface ICanFetchEventsFromStream
     /// <param name="cancellationToken">The <see cref="CancellationToken" />.</param>
     /// <returns>The <see cref="StreamEvent" />.</returns>
     Task<Try<IEnumerable<StreamEvent>>> Fetch(StreamPosition streamPosition, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Gets the current <see cref="StreamPosition"/> offset.
```
jakhog, line 25:
```suggestion
    /// Gets the <see cref="StreamPosition"/> that will be used for the next event written to the stream. 
```

jakhog (changes requested):
Mostly questions, and one suggestion for docs.

Source/EventHorizon/Consumer/Processing/EventsFromEventHorizonFetcher.cs
woksin:
```suggestion
        => new NotImplementedException("GetNextStreamPosition should never be used on this specific fetcher");
```

Source/EventHorizon/Producer/EventHorizon.cs
woksin: It is. This should be okay to do

Source/Events.Store.MongoDB/Streams/StreamFetcher.cs
woksin: This will eventually at some point all go away so I wouldn't bother doing this

Source/EventHorizon/Producer/EventHorizon.cs
woksin: Do you have any name suggestions?

jakhog: Hmm, naming is hard. How about something on the lines of `nextEventPositionInAnyPartition`

woksin: Explicit and mouthy, let's go for that :D

PR #695, opened 2022-05-31
Fixes OutOfMemoryException during public filter validation

woksin:
## Summary

The filter validator for public filters would take the whole event log into memory and compare filtered streams in memory. This could cause OutOfMemoryException if the event log is big enough

### Fixed

- Fixes `OutOfMemoryException` caused by filter validator for public filters.


jakhog (approved):
:shipit:

PR #697, opened 2022-06-03
Event store improvements

woksin:
## Summary

Fixes an issue where Event Store would end up in an endless restarting loop when missing configuration when committing events.

### Fixed

- More friendly errors are logged when getting event store with missing or bad config
- Commits will get a response and not hang whenever the event store fails

### DIFF
--- a/Source/Events.Store/Actors/EventStore.cs
+++ b/Source/Events.Store/Actors/EventStore.cs
@@ -117,7 +117,7 @@ public class EventStore : EventStoreBase
         {
             return respondFailure(_startupFailure);
         }
-        Context.Request(_committer!, request, Context.Sender);
+        Context.Request(_committer!, request!, Context.Sender);
         return Task.CompletedTask;
     }
 
--- a/Source/Events.Store/Actors/EventStore.cs
+++ b/Source/Events.Store/Actors/EventStore.cs
@@ -92,12 +92,7 @@ public class EventStore : EventStoreBase
     /// <inheritdoc />
     public override Task Commit(CommitEventsRequest request, Action<CommitEventsResponse> respond, Action<string> onError)
     {
-        return ForwardToCommitter<CommitEventsRequest, CommitEventsResponse>(request, Respond, RespondWithFailure);
-        Task Respond(CommitEventsResponse response)
-        {
-            respond(response);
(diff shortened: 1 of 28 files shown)


Source/Server/Program.cs
```diff
@@ -42,17 +44,31 @@
...
+    {
+        var config = provider.GetRequiredService<IOptions<TenantsConfiguration>>();
+        if (!config.Value.Any())
+        {
+            logger.LogWarning("No tenants are configured in the Runtime. Without any tenants the Runtime will no function properly.");
+        }
+    }
+    catch (Exception e)
+    {
+        logger.LogWarning(e, "It seems like the Runtime is missing its 'tenants' configuration. Without any tenants the Runtime will no function properly.");
```
mhelleborg, line 58: Error instead of warning?

woksin: Maybe. We could do that

woksin:
```suggestion
        logger.LogError(e, "It seems like the Runtime is missing its 'tenants' configuration. Without any tenants the Runtime will no function properly.");
```

woksin:
Seems like integration tests are failing. Looks like I'll have something fun to debug later :D

Source/Events.Store/Actors/EventStore.cs
```diff
@@ -28,220 +25,128 @@ namespace Dolittle.Runtime.Events.Store.Actors;
...
+        Task RespondWithFailure(Failure failure)
         {
-            _aggregateCommitInFlight.Remove(aggregate);
             respond(new CommitAggregateEventsResponse { Failure = failure });
             return Task.CompletedTask;
         }
     }
 
-    Task<AggregateRootVersion> GetAggregateRootVersion(Aggregate aggregate)
+    Task ForwardToCommitter<TRequest, TResponse>(
```
mhelleborg, line 124: Here you could do a real forward instead of proxying the request.
That would skip sending the response via the eventstore at all

woksin: Ahhh I see. Like this? [6b915ec](https://github.com/dolittle/Runtime/pull/697/commits/6b915eca28982903cd2e7e60ba941a61b53e1be3)

PR #699, opened 2022-06-13
Add support for unified runtime configuration file

woksin:
## Summary

Adds support for a unified json configuration file called `runtime.json` but stil supports the old legacy configuration files under the .dolittle folder. Legacy configurations under the .dolittle folder will overwrite configurations provided in the `runtime.json` configuration file

### Added

- Configuration provider for a `runtime.json` file
- A default `runtime.json` file for default config similar to what we had before under the .dolittle folder

### Deprecated

- Configuration files under the .dolittle folder


woksin:
This should at somepoint be documented on dolittle.io

Source/Configuration/RuntimeConfigurationProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+    public override void Load()
+    {
+        base.Load();
+        foreach (var (key, value) in Data.ToArray())
+        {
+            var newKey = $"{_dolittleConfigSectionRoot}{_delimiter}{key}";
+            Data.Remove(key);
+            Data.Add(newKey, value);
+        }
+    }
```
jakhog, line 37: I thought there was an easy way to add a prefix to a configuration provider in the system, but it seems not - so this might be the best solution right now. However, this `Load()` will not be called again if we enable reload on change - it will only call the `Load(Stream)` signature. So it would be nice if that one was overridden instead of this one.

Source/Configuration/RuntimeConfigurationSource.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+public class RuntimeConfigurationSource : IConfigurationSource
+{
+    /// <inheritdoc /> 
+    public IConfigurationProvider Build(IConfigurationBuilder builder)
+    {
+        var source = new JsonConfigurationSource
+        {
+            Optional = true,
+            Path = "runtime.json",
```
jakhog, line 20: Could we just enable reloading straight away? :)

Source/Server/Program.cs
```diff
@@ -22,14 +24,15 @@
 
 var configBuilder = new ConfigurationBuilder();
 configBuilder.AddJsonFile("appsettings.json");
-configBuilder.AddJsonFile("runtime.json");
-configBuilder.AddLegacyDolittleFiles();
+configBuilder.AddRuntimeFile();
+if (Directory.Exists(".dolittle"))
+{
+    configBuilder.AddLegacyDolittleFiles();
+}
```
jakhog, line 31: Two questions here:

1) Would it make sense to check for existence inside the legacy provider instead of here? That could also make it a little more robust if we check it for every file, instead of the whole directory.

2) As I understand how this code would work - if we update a current running microservice using the legacy files - the default new config in "runtime.json" would be loaded (with dev tenant and localhost mongodb), then all the actual config would be added. Meaning most microservices would end up with a new dev tenant configured on localhost mongodb. Which would probably fail the health checks once started up - and when projections/eventhandlers are registered, they would fail because it cannot connect to the EventStore. Right?

jakhog (changes requested):
Details in comments - but I believe this would break if we deploy it into a microservice using the legacy files for configuration.

Source/Server/Program.cs
woksin: 1) yes we could do that
2) I agree, that sounds about right. One alternative could be to remove the dev tenant from the default config of the production image, and only have it in the dev image. This would be a breaking change. Then we could also start build an additional image which is a "dev-image without mongo", which essentialy would be the same as the current production images

jakhog: Hmm, would be nice if it is not breaking. Could we make up a non-breaking implementation somehow?

woksin: I don't really see how we could since the configurations are added on top of each other and it makes sense for the legacy .dolittle files to override the config in runtime.json

woksin: We could in the legacy file provider override clear everything under 'dolittle:runtime:tenants' if the tenants.json and/or resources.json files are present?

Source/Configuration/RuntimeFileConfigurationProvider.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    void AddOrReplace(string key, string value)
+    {
+        if (Data.TryAdd(key, value))
+        {
+            return;
+        }
+        Data.Remove(key);
+        Data.Add(key, value);
+    }
```
jakhog, line 101: Wouldn't this work?
```suggestion
    void AddAllFromMap(IEnumerable<KeyValuePair<string, string>> map)
    {
        foreach (var (key, value) in map)
        {
            Data[key] = value;
        }
    }
```
Maybe a rename to SetAllFromMap would also be cleaner then.

jakhog (changes requested):
Just a few suggestions.

Source/Configuration/RuntimeFileConfigurationProvider.cs
```diff
@@ -0,0 +1,157 @@
...
+                MapIntoRoot("platform", config);
+                break;
+            case "microservices":
+                MapIntoRoot("microservices", config);
+                break;
+            case "resources":
+                foreach (var key in Data.Keys.Where(_ => _.StartsWith(Constants.CombineWithDolittleConfigRoot("tenants"), StringComparison.InvariantCulture)))
+                {
+                    Data.Remove(key);
+                }
```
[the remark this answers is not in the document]
woksin: Ohh yes that's true. Weird that it didn't crash

Source/Configuration/RuntimeFileConfigurationProvider.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// Initializes a new instance of the <see cref="RuntimeFileConfigurationProvider"/> class.
+    /// </summary>
+    /// <param name="runtimeConfigFile"></param>
+    /// <param name="legacyFilesProvider"></param>
+    /// <param name="buildConfigurationFromFile"></param>
+    public RuntimeFileConfigurationProvider(IFileInfo runtimeConfigFile, IFileProvider? legacyFilesProvider, Func<IConfigurationBuilder, IFileInfo, IConfigurationRoot> buildConfigurationFromFile)
+    {
+        _runtimeConfigFile = runtimeConfigFile;
+        _legacyFilesProvider = legacyFilesProvider;
+        _buildConfigurationFromFile = buildConfigurationFromFile;
```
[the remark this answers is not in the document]
woksin: I added the callback as a way to for instance create more providers for the same config files, for instance for yaml. With the callback that would be quite easy

Source/Server/Server.csproj
```diff
@@ -24,4 +24,12 @@
...
     </ItemGroup>
 
+    <ItemGroup>
+      <Content Include=".dolittle\runtime.yml">
+        <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+      </Content>
+    </ItemGroup>
+
```
jakhog, line 34:
```suggestion
```

Source/Configuration/Configuration.csproj
```diff
@@ -18,6 +18,7 @@
         <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
         <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
         <PackageReference Include="Microsoft.Extensions.FileProviders.Physical" Version="$(MicrosoftExtensionsVersion)" />
-        <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+        <PackageReference Include="NetEscapades.Configuration.Yaml" Version="$(NetEscapadesYaml)" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
jakhog, line 22:
```suggestion
        <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftVersion)" />
```

versions.props
```diff
@@ -20,6 +20,7 @@
         <MongoDBDriverVersion>2.10.1</MongoDBDriverVersion>
         <MongoDBOpenTelemetryVersion>1.0.0</MongoDBOpenTelemetryVersion>
         <MoqVersion>4.13.*</MoqVersion>
+        <NetEscapadesYaml>2.2.0</NetEscapadesYaml>
         <NewtonsoftVersion>12.0.3</NewtonsoftVersion>
```
jakhog, line 24:
```suggestion
        <NewtonsoftVersion>13.0.1</NewtonsoftVersion>
```

jakhog (changes requested):

jakhog (approved):

PR #700, opened 2022-06-16
Add span id to execution context

woksin:
## Summary

To enable tracing of events we need to carry along a span id in the execution context.

### Added

- Span ID to the execution context


Source/Protobuf/ExecutionExtensions.cs
```diff
@@ -27,6 +30,12 @@ public static ExecutionContextContract ToProtobuf(this ExecutionContext executio
             Environment = executionContext.Environment.Value,
             Version = executionContext.Version.ToProtobuf()
         };
+        if (executionContext.SpanId != SpanId.Empty)
+        {
+            var span = new Span<byte>();
+            executionContext.SpanId.Value.CopyTo(span);
+            message.SpanId = ByteString.CopyFrom(span);
+        }
```
woksin, line 38: This is debatable, whether the SpanId field on the ExecutionContext should be nullable or not

Source/Protobuf/ExecutionExtensions.cs
```diff
@@ -44,6 +53,7 @@ public static ExecutionContextContract ToProtobuf(this ExecutionContext executio
             executionContext.Version.ToVersion(),
             executionContext.Environment,
             executionContext.CorrelationId.ToGuid(),
+            executionContext.HasSpanId ? ActivitySpanId.CreateFromBytes(executionContext.SpanId.Span) : SpanId.Empty,
```
woksin, line 56: Comment over would affect this

mhelleborg (approved):

PR #703, opened 2022-07-07
Build results

woksin:
## Summary

Implement Contracts versions 7.4.0 adding Client Build Results to handshake and a management API for getting them

### Added

- Management Grpc API for getting client build results
- Client build results in the handshake request and log them if any


Source/Client/BuildResults.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Dolittle.Runtime.Client;
+
+/// <summary>
+/// Represents a <see cref="ReadOnlyCollection{T}"/> of <see cref="BuildResult"/>.
+/// </summary>
+public class BuildResults
+{
+    public static BuildResults Empty => new(
```
mhelleborg, line 17: This will create it every time. If it's immutable, just do it once?

woksin: Yep. Changed

mhelleborg (approved):

PR #708, opened 2022-08-31
Stream aggregate events

woksin:
## Summary

Implements the new gRPC method for fetching aggregate events in batches that is use by the SDKs to implement more efficient rehydration of aggregate roots.

### Added

- Implemented `FetchForAggregateInBatches` gRPC method that enables SDKs to stream the aggregate events in batches also filtered on event types.


woksin:
Also not to be mentioned in the summary, but there are some internal changes done to adhere to the improved committed aggregate events contracts that has the `CurrentAggregateRootVersion` and `AggregateRootVersion` on each committed aggregate event. That's why there are so many changes because that touched many places in the code, sadly :)

Source/CLI/Runtime/Aggregates/ManagementClient.cs
```diff
@@ -114,9 +114,10 @@ static CommittedAggregateEvents FromProtobuf(EventsContracts.CommittedAggregateE
         => new(
             events.EventSourceId,
             events.AggregateRootId.ToGuid(),
+            events.CurrentAggregateRootVersion,
             events.Events.Select((@event, i) => new CommittedAggregateEvent(
                 new Artifact(events.AggregateRootId.ToGuid(), ArtifactGeneration.First),
-                events.AggregateRootVersion + 1u - (ulong)events.Events.Count + (ulong)i,
+                @event.AggregateRootVersion,
```
woksin, line 120: Just use the aggregate root version on the event

Source/Embeddings.Processing/EmbeddingProcessor.cs
```diff
@@ -174,7 +174,7 @@ async Task<Try> Loop()
             }
 
             var committedEvents = response.Events.ToCommittedEvents();
-            await replaceOrRemoveEmbedding(committedEvents.Last().AggregateRootVersion + 1).ConfigureAwait(false);
+            await replaceOrRemoveEmbedding(committedEvents.AggregateRootVersion).ConfigureAwait(false);
```
woksin, line 177: `CommittedAggregateEvents.AggregateRootVersion` is the current aggregate root version, thus the same as the version of the last event + 1 (I know, we made it this confusing)

Source/Events.Store/CommittedAggregateEventsExtensions.cs
```diff
@@ -18,38 +18,35 @@ public static class CommittedAggregateEventsExtensions
...
     public static Contracts.CommittedAggregateEvents ToProtobuf(this CommittedAggregateEvents committedAggregateEvents)
-    {
-        var aggregateRootVersion = committedAggregateEvents.AsEnumerable().LastOrDefault()?.AggregateRootVersion ?? 0;
-        var protobuf = new Contracts.CommittedAggregateEvents
+        => new()
         {
             AggregateRootId = committedAggregateEvents.AggregateRoot.ToProtobuf(),
             EventSourceId = committedAggregateEvents.EventSource.Value,
-            AggregateRootVersion = aggregateRootVersion
+            AggregateRootVersion = committedAggregateEvents.AggregateRootVersion - 1,
```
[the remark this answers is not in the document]
jakhog: Maybe this is a better place for the comment about the strange behaviour of this version, instead of the `CommittedAggregateEvents.cs` file?

Source/Services/StreamOfBatchedMessagesSender.cs
```diff
@@ -0,0 +1,78 @@
...
+        uint maxBatchSize,
+        IAsyncEnumerator<TDataToBatch> dataEnumerator,
+        Func<TBatch> createNewBatch,
+        Action<TBatch, TDataToBatch> putInBatch,
+        Func<TDataToBatch, TData> convertToData,
+        Func<TBatch, Task> sendBatch)
+    {
+        try
+        {
+            var hasMoreStates = await dataEnumerator.MoveNextAsync().ConfigureAwait(false);
```
jakhog, line 40: As a detail - this could be renamed to `hasMoreData` - this is left over from when it was used for Projections.

Source/Services/StreamOfBatchedMessagesSender.cs
```diff
@@ -0,0 +1,78 @@
...
+        Action<TBatch, TData> putInBatch,
+        Func<TBatch, Task> sendBatch)
+        => Send(maxBatchSize, dataEnumerator, createNewBatch, putInBatch, _ => _, sendBatch);
+    
+    /// <inheritdoc />
+    public async Task Send<TDataToBatch>(
+        uint maxBatchSize,
+        IAsyncEnumerator<TDataToBatch> dataEnumerator,
+        Func<TBatch> createNewBatch,
+        Action<TBatch, TDataToBatch> putInBatch,
```
jakhog, line 34: Maybe this action Should be `Action<TBatch, TData>` we can call the converter below before passing it to this callback.

Source/Events.Store.Services.Grpc/EventStoreGrpcService.cs
```diff
@@ -41,4 +43,51 @@ public override Task<CommitAggregateEventsResponse> CommitForAggregate(CommitAgg
...
+            await responseStream.WriteAsync(response).ConfigureAwait(false);
+            return;
+        }
+
+        await _aggregateEventsBatchSender.Send(
+            MaxBatchMessageSize,
+            fetchResult.Result.EventStream.GetAsyncEnumerator(context.CancellationToken),
+            () => CreateResponse(aggregateRootId, eventSourceId, fetchResult.Result.AggregateRootVersion),
+            (batch, aggregateEvent) => batch.Events.Events.Add(aggregateEvent.ToProtobuf()),
+            aggregateEvent => aggregateEvent.ToProtobuf(),
```
jakhog, line 75: See the comment on the `StreamOfBatchedMessagesSender` below - there are two calls to `aggregateEvent.ToProtobuf()` here - we really only need one.

Source/Events.Store.Services.Grpc/EventStoreGrpcService.cs
```diff
@@ -41,4 +43,51 @@ public override Task<CommitAggregateEventsResponse> CommitForAggregate(CommitAgg
...
+            Events = new Contracts.CommittedAggregateEvents
+            {
+                AggregateRootId = aggregateRootId.ToProtobuf(),
+                EventSourceId = eventSourceId,
+                AggregateRootVersion = currentAggregateRootVersion == AggregateRootVersion.Initial
+                    ? AggregateRootVersion.Initial
+                    : currentAggregateRootVersion - 1,
+                CurrentAggregateRootVersion = currentAggregateRootVersion
+            }
+        };
```
jakhog, line 92: Now we have this conversion and the one in `CommittedAggregateEventsExtensions.cs` - can we remove one of them so it is easier to maintain?

Source/Services/StreamOfBatchedMessagesSender.cs
```diff
@@ -0,0 +1,78 @@
...
+using Google.Protobuf;
+
+namespace Dolittle.Runtime.Services;
+
+/// <summary>
+/// Represents an implementation of <see cref="ISendStreamOfBatchedMessages{TBatch,TData}"/>.
+/// </summary>
+/// <typeparam name="TBatch">The <see cref="Type"/> of the batch message.</typeparam>
+/// <typeparam name="TData">The <see cref="Type"/> of the data to put in a batch.</typeparam>
+public class StreamOfBatchedMessagesSender<TBatch, TData> : ISendStreamOfBatchedMessages<TBatch, TData>
```
jakhog, line 16: Since this thing is really just a class holding a method. Would it be nicer if it was just a static class with a method on it? Not sure if there are like .NET guidelines for these cases - but we don't really need to allocate a class to do this function.

Specifications/Events.Store/for_CommittedAggregateEvents/when_creating_with_a_wrong_artifact.cs
```diff
@@ -14,9 +14,9 @@ public class when_creating_with_a_wrong_artifact : given.events_and_an_artifact
     static CommittedAggregateEvents events;
     static Exception exception;
 
-    Establish context = () => wrong_aggregate_event = new CommittedAggregateEvent(wrong_aggregate_artifact, aggregate_version_after + 1, 3, DateTimeOffset.UtcNow, event_source_id, execution_contexts.create(), event_b_artifact, is_public, "wrong");
+    Establish context = () => wrong_aggregate_event = new CommittedAggregateEvent(wrong_aggregate_artifact, aggregate_version_after, 3, DateTimeOffset.UtcNow, event_source_id, execution_contexts.create(), event_b_artifact, is_public, "wrong");
```
jakhog, line 17: This change makes me wonder - have we change the meaning of a property on the `CommittedAggregateEvent`? I didn't see a change in that class - so I just want to make sure we have checked all the usages of this property in the codebase :)

Specifications/Events.Store/for_CommittedAggregateEventsExtension/given/all_dependencies.cs
```diff
@@ -42,13 +43,18 @@ public class all_dependencies
...
+        {
+            e.AggregateRootVersion = aggregate_version_after_commit++;
+        }
         var result = new Contracts.CommittedAggregateEvents
         {
             AggregateRootId = aggregate_root_id.ToProtobuf(),
             EventSourceId = event_source_id,
             // The aggregate root version of the committed aggregate events is the version of the last event, meaning the aggregate root version before commit happened + number of events - 1
-            AggregateRootVersion = aggregate_version_after_commit - 1
+            AggregateRootVersion = aggregate_version_after_commit - 1,
```
jakhog, line 56: This should also check for the version being 0 before subtracting 1 - right?

Source/Events.Store.Services.Grpc/EventStoreGrpcService.cs
```diff
@@ -17,14 +19,14 @@ namespace Dolittle.Runtime.Events.Store.Services.Grpc;
 [PrivateService]
 public class EventStoreGrpcService : EventStoreBase
 {
+    const uint MaxBatchMessageSize = 2097152; // 2 MB
     readonly IEventStore _eventStore;
-    readonly IFetchCommittedEvents _committedEventsFetcher;
+    readonly StreamOfBatchedMessagesSender<FetchForAggregateResponse, Contracts.CommittedAggregateEvents.Types.CommittedAggregateEvent> _aggregateEventsBatchSender = new();
```
jakhog, line 24: In the Projection service - this is a dependency on the interface. I think this could be a static class instead and then we don't need to do this differently on the two services.

Source/Events.Store.MongoDB/CommittedEventsFetcher.cs
```diff
@@ -83,24 +86,78 @@ public async Task<CommittedEvents> FetchCommittedEvents(ScopeId scope, EventLogS
...
+        ArtifactId aggregateRoot,
+        IEnumerable<Artifact> eventTypes,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await DoFetchForAggregate(
+                eventSource,
+                aggregateRoot,
+                filter => filter & _eventFilter.In(_ => _.Metadata.TypeId, eventTypes.Select(_ => _.Id.Value)),
```
jakhog: Oh - so the filter already only does look at the `TypeId` - then it makes even more sense to only expose the `ArtifactId` on the interface :)

Source/Events.Store.MongoDB/CommittedEventsFetcher.cs
```diff
@@ -83,24 +86,78 @@ public async Task<CommittedEvents> FetchCommittedEvents(ScopeId scope, EventLogS
...
+        ArtifactId aggregateRoot,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await DoFetchForAggregate(
+                eventSource,
+                aggregateRoot,
+                filter => filter,
+                cancellationToken).ConfigureAwait(false);
```
jakhog, line 140: Should we keep track of how many events were sent here as well (like we do in the non-batch method above) and throw the same exception if we don't get all the events?

Source/Events.Store/IFetchCommittedEvents.cs
```diff
@@ -40,6 +42,26 @@ public interface IFetchCommittedEvents
...
+    /// Fetches all <see cref="CommittedAggregateEvent"/>s applied to an Event Source by an Aggregate Root.
+    /// </summary>
+    /// <param name="eventSource">The <see cref="EventSourceId"/> identifying the Event Source.</param>
+    /// <param name="aggregateRoot">The <see cref="ArtifactId"/> identifying the Aggregate Root.</param>
+    /// <param name="eventTypes">The <see cref="IEnumerable{T}"/> of <see cref="Artifact"/> event types of the aggregate events that should be fetched.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken" />.</param>
+    /// <returns>A <see cref="Task" /> that, when resolved, returns the a stream of <see cref="CommittedAggregateEvents"/> containing all <see cref="CommittedAggregateEvent"/> applied to the Event Source by the Aggregate root, in the order of which they appear in the Event Log that have one of the requested event types.</returns>
+    Task<Try<(AggregateRootVersion AggregateRootVersion, IAsyncEnumerable<CommittedAggregateEvent> EventStream)>> FetchStreamForAggregate(EventSourceId eventSource, ArtifactId aggregateRoot, IEnumerable<Artifact> eventTypes, CancellationToken cancellationToken);
```
jakhog: I don't know if this would make the API simpler or worse - but just a thought. These two streaming methods could respond with an `IAsyncEnumerable<CommittedAggregateEvents>` instead of a version + stream of events.

It could either do some batching internally - or just send out a stream with a single event in each. It feels like this could simplify the conversions that now happen outside of this interface a little bit. What do you think?

jakhog (changes requested):

jakhog:
I think it would also be nice if we could spend a little time on writing a few integration tests for the old `FetchForAggregate` API - just to be triply sure we don't break anything, especially with the strange `AggregateRootVersion` number thing.

Source/Events.Store/IFetchCommittedEvents.cs
woksin: I'm not sure. The reason for the `AggregateRootVersion` at least is for knowing the current version of the aggregate root. But we could turn this a bit around as you said and have this return the batches

Specifications/Events.Store/for_CommittedAggregateEventsExtension/given/all_dependencies.cs
woksin: Yes it should :)

Source/Services/StreamOfBatchedMessagesSender.cs
woksin: Agree. I'm too unsure about the guidelines for this

Specifications/Events.Store/for_CommittedAggregateEvents/when_creating_with_a_wrong_artifact.cs
woksin: Yes, I actually added the `AggregateRootVersion` to it. It wasn't there before. We derived the `AggregateRootVersion`  from the last event before.

woksin: Ohh, this is for one event....

woksin: In that case the meaning has not changed, so yeah this is a bit weird

woksin: I've looked at this and the + 1 was just not necessary. This is more correct

Source/Events.Store.MongoDB/CommittedEventsFetcher.cs
woksin: I have no idea how that would work in this case

PR #715, opened 2022-10-04
Fixes a mem leak in AbstractScopedStreamProcessor

mhelleborg:
### Fixed

- Fixes a minor memory leak for scoped stream processors


woksin:
Won't Dispose be called on it either way in the outermost finally-block?

mhelleborg:
> Won't Dispose be called on it either way in the outermost finally-block?

It creates new instances in the inner loop, which needs to be disposed too. This was missing before

woksin (approved):

PR #718, opened 2022-10-26
Add gRPC-web Management service endpoint

woksin:
## Summary

Exposing management services on a new, configurable, grpc-web port so that we can make calls to the management services from a web browser

### Added

- Endpoint configuration for managementWeb with default port 51152
- Exposing management services through grpc-web


jakhog (approved):
:shipit:

PR #726, opened 2023-01-31
Improvements & fixes to fast event processors

mhelleborg:
### Fixed

- Fixes an issue when unregistering eventstore-subscriptions. Was previously missing scopeId on the messages


woksin:
Seems like the 0 ports are problematic?

mhelleborg:
> Seems like the 0 ports are problematic?

Less problematic than the default ports on windows..

PR #740, opened 2023-06-05
Stream Processor retry-rewrite

mhelleborg:
## Summary

Pulled in upstream changes
Added support for partitioned events directly from event log.
Rewritten how retries are done in the actor based event handler processing. Optimizing for lower potential memory footprint, failed events are read from DB when retrying instead of optimizing for IO by keeping them in memory.


TomasEkeli (commented):
i like it!

PR #743, opened 2023-06-15
Add option to not require producer consent when consuming from EventHorizon

mhelleborg:
## Summary

Added the ability to not require producer consent when consuming from event horizon.

To disable required consent, use the following configuration: `DOLITTLE__RUNTIME__EVENTHORIZON__REQUIRECONSENT=false`
This will then pass NoConsent (empty guid) as a valid consent for the consumer, and allow it to consume as normal.

### DIFF
--- a/Source/EventHorizon/Producer/ConsumerService.cs
+++ b/Source/EventHorizon/Producer/ConsumerService.cs
@@ -32,6 +32,7 @@ public class ConsumerService : ConsumerBase
     readonly IConsumerProtocol _protocol;
     readonly IEventHorizons _eventHorizons;
     readonly IMetricsCollector _metrics;
+    readonly IOptions<EventHorizonGlobalSettings> _globalSettings;
     readonly ILogger _logger;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Source/EventHorizon/Producer/ConsumerService.cs
```diff
@@ -93,13 +97,20 @@ public class ConsumerService : ConsumerBase
             arguments.Partition,
             arguments.PublicStream);
 
-        if (!TryGetConsent(arguments, out var consent, out var failureResponse))
+        var consentFound = TryGetConsent(arguments, out var consent, out var failureResponse);
```
TomasEkeli, line 100: i think i'd check the config first, no need to do work about consents if it's going to be ignored

Source/EventHorizon/Producer/ConsumerService.cs
```diff
@@ -114,11 +125,11 @@ public class ConsumerService : ConsumerBase
 
     bool TryGetConsent(ConsumerSubscriptionArguments arguments, out ConsentId consent, out SubscriptionResponse failureResponse)
     {
-        consent = default;
+        consent = ConsentId.NoConsent;
```
TomasEkeli, line 128: much better!

Source/EventHorizon/Producer/ConsumerService.cs
mhelleborg: It does make a difference, as the consent-id is passed to the consumer if it exists. It just does not force it to be there, and will default to returning "NoConsent" instead. It is also quite cheap to check :)

PR #746, opened 2023-06-20
Graceful disconnect

mhelleborg:
## Summary
Adds graceful disconnects for event handlers. This ensures that the system waits for the current handlers to complete before completing shutdowns / disconnects.

Also adresses some minor internal fixes.

### Changed
- Updated projections to accommodate newer MongoDB drivers.
- Rewritten event handlers to support graceful disconnects from the client.

### Fixed
- Resolved an issue in eventlogstream-subscriptions, ensuring correct subscription cancellation.


TomasEkeli:
let's get it out and try it out!

PR #747, opened 2023-06-21
Improve runtime termination procedure

mhelleborg:
## Summary
Adds shutdown hooks to active eventhandlers, ensuring that there are no requests in-flight when the runtime stops.


TomasEkeli:
love it!

